PR #3, opened 2018-02-07
Turn off tracing based on query

michalsanger:
Quick workaround until this is in docker


mrtnzlml:
@michalsanger What is the point here? Mobile team and search are testing this and you want to make the output of every query (except that one) 3x bigger? Why? Apollo Engine implementation is half complete and it doesn't work at this moment. Also, tracing data should not be exposed to the client at all. I am not merging this unless you have very good reason to do so (please help me to understand it)...

michalsanger:
What was the point of releasing the tracing 2 weeks ago? 

This just fix the "too big response" error on Lambda.

mrtnzlml:
> What was the point of releasing the tracing 2 weeks ago?

Yes, that was a mistake.

PR #4, opened 2018-02-09
Prepare all available hotels query to be used with pagination container

tbergq:
Related issue: https://github.com/kiwicom/react-native-app/issues/249

### DIFF
--- a/src/hotel/queries/AllAvailableHotels.js
+++ b/src/hotel/queries/AllAvailableHotels.js
@@ -1,10 +1,12 @@
 // @flow
 
-import { GraphQLNonNull } from 'graphql';
+import { GraphQLNonNull, GraphQLInt } from 'graphql';
 import {
   connectionArgs,
   connectionDefinitions,
   connectionFromArray,
+  cursorToOffset,
+  toGlobalId,
 } from 'graphql-relay';
 
 import { processInputArguments } from '../services/ParametersFormatter';
@@ -43,15 +45,25 @@ export default {
       type: GraphQLAvailableHotelOptionsInput,
     },
     ...connectionArgs,
+    first: {
+      type: GraphQLInt,
+      defaultValue: 50,
+    },
   },
   resolve: async (
     ancestor: mixed,
     args: Object,
     { dataLoader }: GraphqlContextType,
   ) => {
-    const { search: searchArgs } = args;
-    const searchParams = processInputArguments(args);
+    if (args.last) {
+      throw new Error(
+        'Booking.com API does not support querying last, use first query paramter instead.',
+      );
+    }
 
+    const { search: searchArgs } = args;
+    const offset = args.after ? cursorToOffset(args.after) + 1 : 0;
+    const searchParams = processInputArguments({ ...args, offset });
     const availableHotels = await dataLoader.hotel.availabilityByLocation.load(
       searchParams,
     );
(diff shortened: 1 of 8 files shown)


michalsanger:
@tbergq Is this fixing the low number of hotels in list Martin was talking about?

src/hotel/queries/AllAvailableHotels.js
```diff
@@ -50,8 +52,9 @@ export default {
     { dataLoader }: GraphqlContextType,
   ) => {
     const { search: searchArgs } = args;
-    const searchParams = processInputArguments(args);
-
+    const offset = args.after ? cursorToOffset(args.after) + 1 : 0;
+    const rows = args.first || 50;
```
mrtnzlml: 50 should be a default value (`defaultValue`) of the `first` parameter - not an internal value.

src/hotel/queries/AllAvailableHotels.js
```diff
@@ -61,13 +64,28 @@ export default {
...
+        ...args,
+        after: offset,
+      },
     );
 
     return Object.assign(
       {},
       connection,
       { searchParams }, // pass search arguments down
+      // creating page info since booking.com does not provide needed data
```
mrtnzlml: Are you sure about this? This is a responsibility of `connectionFromArray` and it should be quite automatic. What problem are you trying to solve here?

src/hotel/services/ParametersFormatter.js
```diff
@@ -81,7 +81,11 @@ export function prepareRequestParameters(searchParameters: SearchParameters) {
   parameters.currency = searchParameters.currency;
   parameters.min_price = searchParameters.minPrice;
   parameters.max_price = searchParameters.maxPrice;
-  parameters.rows = searchParameters.first || 50;
+  parameters.rows = searchParameters.rows || 50;
```
mrtnzlml: How many times do we need to set 50? This seems like a bug anyway. I would remove this magic internal setting completely. It should be the default value of our queries...

tbergq:
@michalsanger by it self it does not. But this makes query ready to be used with pagination container, so we can load more on scroll to bottom of hotels search.

src/hotel/queries/__tests__/AllAvailableHotels.test.js
```diff
@@ -91,4 +97,35 @@ describe('all hotels query', () => {
...
+          hasPreviousPage
+          startCursor
+          endCursor
+        }
+      }
+    }
+  `;
+    const response = await graphql(query);
+    expect(response.errors[0].message).toEqual(
+      'Booking.com API does not support querying last, use first query paramter instead',
```
mrtnzlml: I would move this test to the separate file because you don't need to REST API mocks.

src/hotel/queries/AllAvailableHotels.js
```diff
@@ -61,13 +73,28 @@ export default {
...
+      {
+        pageInfo: {
+          startCursor: toGlobalId('arrayconnection', offset.toString()),
+          endCursor: toGlobalId(
+            'arrayconnection',
+            (offset + args.first - 1).toString(),
+          ),
+          hasPreviousPage: offset > 0,
+          hasNextPage: connection.edges.length === args.first,
```
mrtnzlml: I don't understand this. Shouldn't this be almost always `true`? We have almost infinite scroll because we don't know where is the end...

src/hotel/queries/AllAvailableHotels.js
```diff
@@ -43,15 +45,25 @@ export default {
...
   resolve: async (
     ancestor: mixed,
     args: Object,
     { dataLoader }: GraphqlContextType,
   ) => {
-    const { search: searchArgs } = args;
-    const searchParams = processInputArguments(args);
+    if (args.last) {
+      throw new Error(
+        'Booking.com API does not support querying last, use first query paramter instead',
```
mrtnzlml:
```diff
- 'Booking.com API does not support querying last, use first query paramter instead',
+ 'Booking.com API does not support querying last, use first query parameter instead.',
```

src/hotel/queries/AllAvailableHotels.js
tbergq: the idea here is that we ask for 50, if we get 50, there is most likely another page. If we ask for 50, but get less, then there are no more data to be loaded.

mrtnzlml: Ah, understand. I need to comment this resolver properly. It's quite complicated due to all assumptions and limitations involved...

PR #9, opened 2018-02-25
Redirect to GraphiQL on production

### LINKED ISSUE #7: GET request fails with with misleading error
The error:
```
{
  "message": "Missing Authentication Token"
}
```

Better to forward GET requests to https://kiwi-graphiql.now.sh

michalsanger:
close #7

### DIFF
--- a/src/graphqlServer.js
+++ b/src/graphqlServer.js
@@ -25,6 +25,9 @@ const app = express();
 app.use(cors({ methods: ['GET', 'POST'] }));
 
 app.use('/', (request: $Request, response: $Response) => {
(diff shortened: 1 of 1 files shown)


mrtnzlml:
We should IMHO allow GET and/or add instructions how to use it instead of redirecting away. This is such a bad behavior of API endpoint... 😞 Also please feel free to push it directly next time (you are a merge master here).

michalsanger:
Different solution should be discussed in the issue. Proposed redirect was without any comment for few days so I thought everyone agree with this. Feel free to create new issue and suggest different solution.

I'd rather let every change go through code review.

PR #18, opened 2018-03-05
Query for fetching location by ID

### LINKED ISSUE #11: Support fetching location by id
User story: When returning user comes to the website, we want him to be able to be able to perform search from the previous search history in place picker. We also want user to be able in the future mark certain flights as favorites and show them in a special menu for the user.

This functionality requires us to save `locationId`s in the locals storage and load locations based on those IDs from API. For this call we can use `locationId` parameter in the `allLocations` call or implement it as a separate call, what makes more sense for the current state of graphql server.

michalsanger:
close #11

The implementation points to some test refactoring needs. The refactoring is done in own PR to keep this PR small.

### DIFF
--- a/src/location/dataloaders/Location.js
+++ b/src/location/dataloaders/Location.js
@@ -1,14 +1,27 @@
 // @flow
 
+import DataLoader from 'dataloader';
+import stringify from 'json-stable-stringify';
+
 import LocationSuggestions from './LocationSuggestions';
+import { batchGetLocations } from './Fetcher';
 
 import type { Location, Options } from '../Location';
 
 export default class LocationDataLoader {
   locationSuggestionsDataLoader: LocationSuggestions;
+  dataLoader: DataLoader<Object, Location[] | Error>;
 
   constructor(dataloader: LocationSuggestions) {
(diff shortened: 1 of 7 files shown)


src/location/dataloaders/Location.js
```diff
@@ -1,14 +1,25 @@
...
+  dataLoader: DataLoader<Object, Location[] | Error>;
 
   constructor(dataloader: LocationSuggestions) {
     this.locationSuggestionsDataLoader = dataloader;
+    this.dataLoader = new DataLoader(
+      (urlParameters: $ReadOnlyArray<Object>) => {
+        return batchGetLocations(urlParameters);
+      },
+      {
+        cacheKeyFn: key => JSON.stringify(key),
```
mrtnzlml: Please use [`json-stable-stringify`](https://github.com/kiwicom/graphql/blob/2070797b6fcea2f17fe04e42df8c5521a0a0cdb7/src/hotel/dataloaders/HotelByID.js#L38). This is fragile.

michalsanger:
@mrtnzlml all fixed

Issue #17, opened 2018-03-05
Locations refactoring


michalsanger:
- [x] add opaque ID
- [x] enable opaque ID as input in Location dataloder
- [x] `src/flight/types/outputs/RouteStop.js` should call `location.loadById`
- [x] remove `load` and `loadMany` from `src/location/dataloaders/Location.js`
- [x] rename `LocationSuggestions` -> `Locations`
- [x] rename methods in LocationSuggestions:
	`load` -> `loadAll`
	`loadByKey` -> `loadByTerm`
	`loadMany` -> `loadByTerms`

mrtnzlml:
@michalsanger Are you going to continue with this or we can close it?

michalsanger:
@mrtnzlml keep it here please, it's still valid. It doesn't have to be me who implements it

mrtnzlml:
Ok. I don't believe someone else will actually implement it. It's too ambitious with unclear intentions... 🙂 Hopefully, I am wrong.

jaroslav-kubicek:
Would be also nice to remove the necessity to pass `locale` from resolvers and instead provide it when dataloader is created. See FAQ* dataloaders for example.

The only question that remains is whether we want to support a deprecated way of specifying the locale - via query args.

jaroslav-kubicek:
ok, that's true, it's already used. Then ignore my comment for now, it would just complicate things now.

mrtnzlml:
Well, but the first part is absolutely valid - we have to refactor it to use the HTTP header because we are now sending only the header I think. It's now way more important than renaming loader methods...

mrtnzlml:
@radkag This can be closed, right? 🙂

radkag:
@mrtnzlml yes, i was thinking why is it still open. :))

mrtnzlml:
@radkag I am really confused by this development style - seems like it's not fixed yet: https://github.com/kiwicom/graphql/pull/187?

radkag:
@mrtnzlml Yes, I should not close it.

PR #19, opened 2018-03-05
Location tests refactoring

michalsanger:
Merge after #18 

There were some untestable lines because Dataloader throws it's own exception when returned value is Error. 

There is more refactoring to be done, see #17

### DIFF
--- a/jest.json
+++ b/jest.json
@@ -10,7 +10,7 @@
   ],
   "coverageThreshold": {
     "global": {
-      "branches": 79,
+      "branches": 80,
       "functions": 80,
       "lines": 80,
       "statements": 80
--- a/jest.json
+++ b/jest.json
@@ -10,7 +10,7 @@
   ],
   "coverageThreshold": {
     "global": {
-      "branches": 80,
+      "branches": 79,
       "functions": 80,
       "lines": 80,
       "statements": 80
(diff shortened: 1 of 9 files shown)


mrtnzlml:
> There were some untestable lines because Dataloader throws it's own exception when returned value is Error.

I don't really understand what was the problem, but there is also optimistic data loader if you want to work with errors returned in batch loads.

michalsanger:
The problem is when the return type of batchLoad function is like this `Promise<Array<Location[] | Error>>` then Flow force you to have this to handle the `Error` after calling `dataloader.load(...)`:
```js
if (possibleValues instanceof Error) {
  throw possibleValues;
}
```

But this case never happen. If the batchLoad returns `Error` dataloader rejects the promise and with `async/await` the `Error` is thrown. Those untested lines in `yarn test-ci` pissed me so much I created this PR :-) Plus I added tests for the new `location` query.

mrtnzlml:
> Those untested lines in yarn test-ci pissed me so much I created this PR...

I know you like to hear something clever from more clever people so here is one wisdom from Dan Abramov for you: https://twitter.com/dan_abramov/status/944939741094535168 😊

michalsanger:
@mrtnzlml Thanks for the link. Please also give me name of medicine you are now using to handle your strong OCD

PR #24, opened 2018-03-08
set up CI step for triggering pipeline

### LINKED ISSUE #2: Move from Serverless to Docker environment
We are responding with huge JSONs especially because of tracing data (>6MB). But this is limit of AWS Lambda. We need to be able to send much more if needed and not just fail. Docker will solve this and it will simplify our stack (it should contain Apollo Engine as well)...

Fallion:
Closes #2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,11 @@ jobs:
       - attach_workspace:
           at: .
       - run: sh ./scripts/deploy.sh
+  rancher deploy:
+    docker:
+      - image: pstauffer/curl:latest
+    steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -45,6 +45,12 @@ jobs:
       - attach_workspace:
           at: .
       - run: sh ./scripts/deploy.sh
+  rancher deploy:
+    docker:
+      - image: pstauffer/curl:latest
+    steps:
+      - checkout
```
mrtnzlml: @Fallion Is the checkout step necessary?

Fallion: Nope. Sorry. Fixed.

PR #31, opened 2018-03-10
Fetch by slug and radius

### LINKED ISSUE #16: Support fetching by slug and radius
User story: when you enter on the site by clicking on specific search URL, user wants to load a form with pre-filled locations and map with rendered data points as fast as possible. 

GraphQL should therefore support fetching all locations from slug and radius (which we get from this URL for example: https://www.kiwi.com/en/search/bratislava-slovakia-169km/56.25-19.42-250km) in order to support that. For now we filter them by cities only, it is possible that we would use stations/airports in the future as well for specific cases.

michalsanger:
close #16

### DIFF
--- a/src/location/services/SlugRadiusParser.js
+++ b/src/location/services/SlugRadiusParser.js
@@ -1,0 +1,62 @@
+// @flow
+
+type SlugRadius = {|
+  slug: string,
+  radius: number,
+|};
+type GpsRadius = {|
+  gps: { lat: number, lng: number },
+  radius: number,
+|};
+
+export default function parse(slugRadius: string): SlugRadius | GpsRadius {
+  const radius = getRadius(slugRadius);
+  const slug = stripRadius(slugRadius, radius);
+
+  // slug is not in lat-lng format
+  if (slug.match(/^[0-9.-]+$/g) === null) {
+    return { slug, radius };
+  }
+
+  const gps = getGps(slug);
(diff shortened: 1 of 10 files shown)


michalsanger:
May I ask for review plz? @jalbertsr

src/location/services/SlugRadiusParser.js
```diff
@@ -0,0 +1,62 @@
...
+  radius: number,
+|};
+type GpsRadius = {|
+  gps: { lat: number, lng: number },
+  radius: number,
+|};
+
+export default function parse(slugRadius: string): SlugRadius | GpsRadius {
+  const radius = getRadius(slugRadius);
+  let slug = stripRadius(slugRadius, radius);
```
jalbertsr: why not `const` ?

michalsanger: Good catch, fixed.

tbergq: What about adding this to eslint?-
`eslint prefer-const` -> `This rule is aimed at flagging variables that are declared using let keyword, but never reassigned after the initial assignment.`

michalsanger: @tbergq good idea, could you make PR with that addition?

tbergq: Sure

PR #33, opened 2018-03-13
Add room size to room output

tbergq:


src/hotel/dataloaders/HotelRooms.js
```diff
@@ -71,32 +71,9 @@ export default class HotelRoomBlocksDataloader {
...
           return [];
         }
-        return this.sanitizeHotelRooms(rooms.room_data, hotelId);
+        return sanitizeHotelRooms(rooms.room_data, hotelId);
       });
       return hotelRooms.reduce((a, b) => a.concat(b));
     });
   }
-
-  sanitizeHotelRooms(rooms: Object, hotelId: number): HotelRoomType[] {
```
mrtnzlml: I just hope that they have the same rooms format for both responses (both data-loaders)... :)

tbergq: The two functions were identical, and there is nothing better than duplicated code :)

PR #41, opened 2018-03-19
allow tracing for non-lambda servers

### LINKED ISSUE #36: Proxy: tracing doesn't work
There is this error in AE:

> This trace information is incomplete. Check the product documentation to properly enable tracing in your GraphQL server and ensure you the latest version of the Engine Proxy is used.

Tracing doesn't work (only timing) and as a consequence, we cannot see error details:

screen shot 2018-03-14 at 16 51 58

I am now in the situation where one client fails but I don't know why and where to start.

Fallion:
Closes #36

### DIFF
--- a/src/graphqlServer.js
+++ b/src/graphqlServer.js
@@ -34,7 +34,7 @@ app.use('/', (request: $Request, response: $Response) => {
 
   const token = request.header('authorization') || null;
   const context = createContext(token);
-  if (process.env.NODE_ENV !== 'test') {
+  if (process.env.NODE_ENV !== 'test' && !process.env.IS_LAMBDA) {
     const traceCollector = new TraceCollector();
     traceCollector.requestDidStart();
     context._traceCollector = traceCollector;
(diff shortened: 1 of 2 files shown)


Fallion:
screen shot 2018-03-19 at 19 42 50

src/graphqlServer.js
```diff
@@ -67,8 +67,7 @@ function createGraphqlServer(schema, context) {
     },
     extensions: () => {
       const traceCollector = context._traceCollector;
-      if (true || !traceCollector) return {}; // eslint-disable-line no-constant-condition
-      // Quick hack, remove 'true ||' once we can handle huge tracing output
+      if (!!process.env.IS_LAMBDA || !traceCollector) return {};
```
michalsanger: Isn't it better not to setup the tracing on Lambda at all? So put the `process.env.IS_LAMBDA` check on line 37

Fallion: Good point. Didn't see that, will update now.

Fallion: Fixed.

PR #46, opened 2018-03-20
Pass language parameter to get room description

tbergq:
Language parameter was not passed down to get room description translated.

### DIFF
--- a/src/hotel/dataloaders/HotelByID.js
+++ b/src/hotel/dataloaders/HotelByID.js
@@ -29,8 +29,10 @@ export default new OptimisticDataloader(
     const hotels = hotelIds.map(id => {
       const hotelData = response.result.find(h => h.hotel_id == id);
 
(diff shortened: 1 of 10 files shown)


src/hotel/dataloaders/HotelByID.js
```diff
@@ -29,8 +29,10 @@ export default new OptimisticDataloader(
     const hotels = hotelIds.map(id => {
       const hotelData = response.result.find(h => h.hotel_id == id);
 
-      if (!hotelData) return new Error('Requested hotel does not exist.');
-      return sanitizeHotel(hotelData);
+      if (!hotelData) {
+        return new Error('Requested hotel does not exist.');
+      }
+      return sanitizeHotel(hotelData, language || '');
```
michalsanger: Shouldn't the `language` param be optional? What's the point of `''` as fallback?

tbergq:
@michalsanger Fixed flow types, and removed fallback to empty string. Please check again.

PR #54, opened 2018-04-04
Add "parentCategoryId" argument to allFAQCategories query

jaroslav-kubicek:
In order to enable fetching just a subset of FAQ categories that belong to one specific category, I'm introducing new `parentCategoryId` argument to the query `allFAQCategories`.

BTW - I have one unresolved question - should it return empty list as it's now or error if the category with such id doesn't exist?


michalsanger:
I'd return an `error`

jaroslav-kubicek:
updated

jaroslav-kubicek:
god dammit, I want something little bit different 🙃

PR #55, opened 2018-04-04
Add new "FAQCategory" query that enables fetching just a single category

jaroslav-kubicek:
Ok, this is what I really wanted... 🙂 

Changes:
- new `FAQCategory` query to fetch just single FAQ category, possibly with its subcategories and articles
- mischievously doing breaking change - type name `AllFAQCategories` changed to `FAQCategory` because the previous naming was just bad. This should keep consistency with `HotelCity`, `Location` etc. as well.

### DIFF
--- a/src/FAQ/queries/FAQCategory.js
+++ b/src/FAQ/queries/FAQCategory.js
@@ -1,0 +1,67 @@
+// @flow
+
+import { GraphQLID, GraphQLNonNull } from 'graphql';
+import { fromGlobalId } from 'graphql-relay';
+
+import type { GraphqlContextType } from '../../common/services/GraphqlContext';
+import FAQCategory, {
+  type FAQCategoryType,
+} from '../types/outputs/FAQCategory';
+import LanguageInput from '../../common/types/inputs/LanguageInput';
+
+const findCategory = (
+  categories: FAQCategoryType[],
+  categoryId: number,
+): FAQCategoryType | null => {
+  const parentCategory = categories.find(c => c.id === categoryId);
+
+  if (parentCategory) {
+    return parentCategory;
+  }
+
+  for (const category of categories) {
+    const subcategories = category.subcategories || [];
+    const subcategory = findCategory(subcategories, categoryId);
+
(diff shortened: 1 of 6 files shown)


src/FAQ/queries/FAQCategory.js
```diff
@@ -0,0 +1,72 @@
+// @flow
+
+import { GraphQLID, GraphQLNonNull } from 'graphql';
+import { fromGlobalId } from 'graphql-relay';
+
+import type { GraphqlContextType } from '../../common/services/GraphqlContext';
+import FAQCategory, {
+  type FAQCategoryType,
+} from './../types/outputs/FAQCategory';
```
michalsanger: `./../` looks strange, `../` is enough

src/FAQ/queries/FAQCategory.js
```diff
@@ -0,0 +1,72 @@
...
+  categoryId: number,
+): FAQCategoryType | null => {
+  const parentCategory = categories.find(c => c.id === categoryId);
+
+  if (parentCategory) {
+    return parentCategory;
+  }
+
+  for (const category of categories) {
+    if (!category.subcategories) {
```
michalsanger: How about being a little bit more elegant?
```js
const subcategories = category.subcategories || [];
return findCategory(subcategories, categoryId);
```

src/FAQ/queries/FAQCategory.js
```diff
@@ -0,0 +1,72 @@
...
+      description:
+        'Language in which the titles and perexes of FAQ categories are returned.',
+    },
+  },
+  resolve: async (
+    ancestor: mixed,
+    { language, id }: Object,
+    { dataLoader }: GraphqlContextType,
+  ) => {
+    const idObject = fromGlobalId(id);
```
michalsanger: again one more subjective elegant note:
```js
const categoryId = Number(fromGlobalId(id).id);
```

jaroslav-kubicek:
fixed, cc @michalsanger

PR #59, opened 2018-04-06
Unify output types in all current FAQ queries

jaroslav-kubicek:
⚠️  *Breaking changes* on sight - hopefully, no one cares too much as it's related only to FAQ newborn project.

`allFAQs` query output has no more single `results` field with custom subfields but uses `FAQArticle` type.

instead of this:
```
query SearchAllFAQsQuery($search: String, $language: Language) {
   allFAQs(search: $search, language: $language) {
      edges {
        node {
          results {
            id
            title
            articleId
            content
          }
        }
      }
    }
  }
```

it's now this:
```
query SearchAllFAQsQuery($search: String, $language: Language) {
  allFAQs(search: $search, language: $language) {
    edges {
      node {
        id
        title
        perex
      }
    }
  }
}
```


src/FAQ/types/FAQArticle.js
```diff
@@ -22,6 +22,20 @@ export type FAQArticleDetail = {
   perex: string,
   upvotes: number,
   downvotes: number,
+  language: string,
+};
+
+// this is sanitized output from "/search?q=term" search endpoint
+export type SanitizedArticleFromSearch = {
```
michalsanger: This type name is really verbose :-( Isn't there little bit mess with these types? If there is `FAQArticleDetail` I would expect also `FAQArticle` exists. Also why `SanitizedFAQArticleItem` isn't called just `FAQArticleItem`?

src/FAQ/dataloaders/searchFAQ.js
```diff
@@ -3,30 +3,55 @@
...
 import Config from '../../../config/application';
-import { sanitizeFAQ } from './ApiSanitizer';
+import type { SanitizedArticleFromSearch } from '../types/FAQArticle';
 
 export type Args = {
   search: string,
   language: string,
 };
 
+export type ApiResponse = {|
```
michalsanger: Why is it exported?

jaroslav-kubicek: mistake

src/FAQ/types/FAQArticle.js
jaroslav-kubicek: ok, all these types are related to dataloaders, hence I dissolved this file completely and moved all types to their related dataloaders. Now it should be much more clear.

jaroslav-kubicek:
updated, cc @michalsanger  (but pls no merge now before demo :D)

src/FAQ/types/FAQArticle.js
mrtnzlml: 📜 https://github.com/kiwicom/graphql#data-loaders

jaroslav-kubicek: @mrtnzlml  so you suggest to include error / not found types as well with this comment?

jaroslav-kubicek: ok, thanks, it gave me the motivation to add `NotFound` scenario for FAQ article. Now we can hopefully merge it. 🙏

src/FAQ/dataloaders/getFAQArticle.js
```diff
@@ -3,7 +3,20 @@
...
+export type FAQArticleDetail = {
+  id: number,
+  content: string,
+  title: string,
+  perex: string,
+  upvotes: number,
+  downvotes: number,
+  language: string,
+};
+type APIResponse = NotFound | FAQArticleDetail;
```
michalsanger: What's the meaning of this type? `NotFound` is never used

src/FAQ/dataloaders/getFAQArticle.js
```diff
@@ -14,9 +27,19 @@ const getFAQArticle = async (
   originalId: string,
   language: string,
 ): Promise<FAQArticleDetail> => {
-  return get(Config.restApiEndpoint.FAQArticle(originalId), null, {
+  const url = Config.restApiEndpoint.FAQArticle(originalId);
+  const article: APIResponse = await get(url, null, {
```
michalsanger: Is the response of `get` really the `APIResponse`? Then why the 
```js
return {
  ...article,
  language,
};
```

src/FAQ/dataloaders/getFAQArticle.js
```diff
@@ -3,7 +3,20 @@
 import Dataloader from 'dataloader/index';
 import { get } from '../../common/services/HttpRequest';
 import Config from '../../../config/application';
-import type { FAQArticleDetail } from '../types/FAQArticle';
+
+type NotFound = {
+  detail: string,
+};
+export type FAQArticleDetail = {
```
michalsanger: Btw why don't you use exact types?

src/FAQ/dataloaders/getFAQArticle.js
jaroslav-kubicek: fixed

src/FAQ/dataloaders/getFAQArticle.js
jaroslav-kubicek: it's used for condition - to make sure it can be potentially resolved with object, but with object holding just `{ detail: string }`
```
if (!article || !article.content) {
```

src/FAQ/dataloaders/getFAQArticle.js
jaroslav-kubicek: it's not possible to use them everywhere, doesn't work with 1) spreading 2) intersection types `Type1 & Type2`

jaroslav-kubicek: but I introduced them whenever possible

src/FAQ/dataloaders/getFAQArticle.js
michalsanger: Don't get it. `article.content` is not part of `NotFound` type. `!article || article.detail` would make sense to me.

jaroslav-kubicek: ok, is it fine now? @michalsanger

PR #60, opened 2018-04-09
Add "FAQArticle" query

jaroslav-kubicek:
Hey, I'm adding `FAQArticle` query as that should be the last missing piece of puzzle to fully support FAQ in GraphQL now  🎉

### DIFF
--- a/src/FAQ/queries/FAQArticle.js
+++ b/src/FAQ/queries/FAQArticle.js
@@ -1,0 +1,43 @@
+// @flow
+
+import { GraphQLID, GraphQLNonNull } from 'graphql';
+import { fromGlobalId } from 'graphql-relay';
+
+import FAQArticle from '../types/outputs/FAQArticle';
+import LanguageInput from '../../common/types/inputs/LanguageInput';
+import type { GraphqlContextType } from '../../common/services/GraphqlContext';
+
+export default {
+  type: FAQArticle,
+  description: 'Retrieve single FAQ article.',
+  args: {
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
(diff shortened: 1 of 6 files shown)


src/FAQ/queries/FAQArticle.js
```diff
@@ -0,0 +1,39 @@
...
+  },
+  resolve: async (
+    ancestor: mixed,
+    { language, id }: Object,
+    { dataLoader }: GraphqlContextType,
+  ) => {
+    const { id: originalId, type } = fromGlobalId(id);
+
+    if (type !== 'FAQArticle') {
+      throw new Error('This id does not belong to any FAQArticle.');
```
michalsanger:
```js
throw new Error(
  `FAQArticle ID mishmash. You cannot query FAQ with ID ` +
  `'${id}' because this ID is not ID of the FAQArticle. ` +
  `Please use opaque ID of the FAQArticle.`,
);
```

src/FAQ/types/outputs/FAQArticle.js
```diff
@@ -43,6 +43,11 @@ export default new GraphQLObjectType({
         args: Object,
         { dataLoader }: GraphqlContextType,
       ): Promise<string> => {
+        // $FlowExpectedError: "content" is present only for FAQArticleDetail
```
michalsanger: No need for such optimization. Dataloader implements cache, so feel free to `await dataLoader.FAQArticle.load` and don't worry it will call another API request. Comment explaining why `content` can be null is helpful.

jaroslav-kubicek:
all should be fixed, cc @michalsanger

PR #66, opened 2018-04-18
geoip: Add

oreqizer:
Added a root query **geoIP** that returns geography info of an IP address.

Used by Kiwi's landing page.


src/common/services/JsonFetcher.js
```diff
@@ -8,7 +8,7 @@ export async function fetchJson(
   url: string,
   method: string = 'GET',
   options?: Object = {},
-): Promise<Object | Array<Object>> {
```
oreqizer: This was impossible to be well overwritten by the `GeoIP` type

oreqizer: screen shot 2018-04-18 at 11 10 49

using `Object` forces me to use inexact types

src/geoip/types/outputs/GeoIPType.js
```diff
@@ -0,0 +1,28 @@
...
+      description: 'ISO country code',
+    },
+    latitude: {
+      type: GraphQLFloat,
+      description: 'Latitude',
+    },
+    longitude: {
+      type: GraphQLFloat,
+      description: 'Longitude',
+    },
```
mrtnzlml: We have `Coordinates` type for this. Maybe you want to use it?

src/geoip/queries/geoIP.js
```diff
@@ -0,0 +1,18 @@
...
+
+import GeoIPType from '../types/outputs/GeoIPType';
+import getGeoIP from '../apis/geoip';
+
+const geoIPQuery = {
+  type: GeoIPType,
+  description: 'Geography info by an IP address',
+  args: {
+    ip: {
+      type: new GraphQLNonNull(GraphQLString),
```
mrtnzlml: Type `string` is not a very good type for IP (see: `geoIP(ip: "x")`). The correct way would be to create (or use already created) IP input types. Would you like to improve it?

oreqizer:
Resolved two things, third is awaiting response.

src/common/types/scalars/IP.js
```diff
@@ -0,0 +1,20 @@
+import { GraphQLScalarType, Kind } from 'graphql';
+import isIP from 'validator/lib/isIP';
+
+// Exported for testing
+export const maybeIP4 = val => (isIP(val, 4) ? val : null);
```
mrtnzlml: The only reason why I wanted custom scalar type is the fact that you can do quick validations and improve DX. Therefore you should throw an exception here. Now it's still possible to call `geoIP` query with whatever you want and the result is this call: `https://geoip-api-prod.skypicker.com/geoip-api?ip=null` (GeoIP server cannot handle it).

This is what you'll get with `Error` instead of `null`:

```json
{
  "errors": [
    {
      "message": "Expected type IP!, found \"x\"; Value is not a valid IP address",
      "locations": [
        {
          "line": 2,
          "column": 13
        }
      ]
    }
  ],
}
```

And this is what you'll get with current code:

```json
{
  "errors": [
    {
      "message": "Bad Gateway",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "geoIP"
      ],
      "_proxy": {
        "statusCode": 502,
        "url": "https://geoip-api-prod.skypicker.com/geoip-api?ip=null"
      }
    }
  ],
  "data": {
    "geoIP": null
  },
}
```

src/common/types/scalars/IP.js
```diff
@@ -0,0 +1,20 @@
+import { GraphQLScalarType, Kind } from 'graphql';
```
mrtnzlml: Please rebase on top of the master branch and *squash* everything. You'll have to fix eslint error here...

src/common/types/scalars/IP.js
```diff
@@ -0,0 +1,31 @@
...
+// Exported for testing
+export function maybeIP4(val: mixed) {
+  if (isIP(val, 4)) {
+    return val;
+  }
+
+  throw new Error(
+    `Expected type IP!, found "${String(
+      val,
+    )}"; Value is not a valid IP address`,
```
mrtnzlml: @oreqizer Just a note (I will fix it for you) - this actually generates this message:

```
"message": "Expected type IP!, found \"x\"; Expected type IP!, found \"x\"; Value is not a valid IP address",
```

So the correct error value is just this: `Error('Value is not a valid IP address')`... 🙂

oreqizer: ah okay, good to know, thanks

mrtnzlml:
Integrated, thanks.

PR #72, opened 2018-04-20
eslint rule for connectionFromArray import restriction, resolves #53

### LINKED ISSUE #53: Use custom connectionFromArray everywhere
> Could you please use [this Eslint rule](https://eslint.org/docs/rules/no-restricted-imports) to make sure we'll use it everywhere even in the future? See for example this: https://github.com/kiwicom/react-native-app/blob/e39ea6a773dfffc14180edce4cfb7e1caa4b0b71/.eslintrc.js#L40-L74

Rulsky:
This is a PR for a #53 with addition of eslint rule to restrict importing native function `connectionFromArray` from `graphql-relay` to encourage usage of in-house implementation.


package.json
```diff
@@ -24,13 +24,13 @@
   "devDependencies": {
     "babel-cli": "^6.24.1",
     "babel-core": "^6.25.0",
-    "babel-eslint": "^7.2.3",
+    "babel-eslint": "~8.2.2",
```
mrtnzlml: Why did you decide to change it from `^` to `~`? Is there a good reason for that? Can we please stick with the [caret range](https://yarnpkg.com/lang/en/docs/dependency-versions/#toc-caret-ranges)?

Rulsky: I copied exact version from kiwicom/react-native-app. Will fix it.

Rulsky:
Changed a version range for babel-eslint

package.json
mrtnzlml: Haha, now the question is why do we have it in our RN repo. Thanks for pointing this out... :)

mrtnzlml:
@Rulsky You stated [in this comment](https://github.com/kiwicom/graphql/issues/53#issuecomment-383059779) that there should be few changes in other files but they are not part of this PR. But this is basically what is this PR about so you have to fix this as well otherwise the tests are going to fail (and therefore we cannot merge it)... 🙂

mrtnzlml:
Also, we prefer clean git history of PRs at any time so you should squash your changes into one commit and rebase it on top of the master branch. However, if you don't know how to do that I can fix it for you... 😉

Rulsky:
updated consumers of `connectionFromArray`, squashed into one commit, CI passed.

mrtnzlml:
@Rulsky Thank you very much! :)

Rulsky:
Thanks for the opportunity

PR #74, opened 2018-04-20
Add locale to booking queries

jaroslav-kubicek:
It also enables to pass both Booking.id and Booking.databaseId to `booking(id: ID!)` query


jaroslav-kubicek:
> No. Evolve not copy-paste... We should use proper methods like Accept-Language header and not custom parameters for this. Locale input type should be completely deprecated...

Ok, I had no idea about this, I'm always trying to comply with already existing queries. So let's include this into contribution guide coming soon 🙂

jaroslav-kubicek:
Closing in favor of #82

PR #78, opened 2018-04-21
Add isReturn field to Leg type

### LINKED ISSUE #77: Add "isReturn" field to Leg type
Make possible to detect whether specific Leg belongs to return flight or not. Booking API already returns this as `return` with values 1 or 0. We just need to return this information too, as boolean.

jaroslav-kubicek:
Closes #77 

I'm adding `isReturn` field to `Leg` to determine whether specific Leg is related to return flight or not.


src/booking/dataloaders/__tests__/__datasets__/booking.withReturn.json
```diff
@@ -0,0 +1,326 @@
...
+  ],
+  "route": [
+    "PRG",
+    "NRT"
+  ],
+  "sandbox": 0,
+  "segments": null,
+  "status": "confirmed",
+  "transfer_ids": []
+}
```
mrtnzlml: It's little bit silly to use 326 lines of something just to test one property, isn't it?

PR #82, opened 2018-04-22
Use "Accept-Language" header for localization

### LINKED ISSUE #79: Use Accept-Language header for localization
See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

Inspiration: https://www.yelp.ca/developers/graphql/guides/localization

Also, please deprecate `Locale` input type.

jaroslav-kubicek:
Closes #79 

I'm kicking off localization again, by using "Accept-Language" header and storing the value inside the context.

Few points
- Right now, I accept locales with `-` not with `_`, e.g. `cs-CZ`, since it's used in REST APIs, what do we want to use, `cs_CZ`?
- old `Locale` input params still take precedence over header - to keep backward compatibility 
- `en-US` is the default


jaroslav-kubicek:
There is no standard way how to decprecate arguments, except enum values individualy. Any proposals?

jaroslav-kubicek:
Ok, updated, cc @mrtnzlml 

- `en_US` default (with `_`)
- all Locale values deprecated + description updated

src/flight/dataloaders/BookingUrlBuilder.js
```diff
@@ -25,7 +25,8 @@ export async function buildBookingUrl(
     price: priceInEur,
     token,
   });
-  const urlLocale = locale && LocaleMap[locale] ? LocaleMap[locale] : 'en';
+  const urlLocale =
+    typeof locale === 'string' ? LocaleMap[locale.replace('_', '-')] : 'en';
```
mrtnzlml: @jaroslav-kubicek Are you sure that every language we accept is actually accepted by our booking API? This is actually one of our pains in React Native app - we can use whatever language we want but it doesn't work because Booking.com works only with a limited subset. However, we can probably fix this alter...

jaroslav-kubicek: no, it doesn't work with everything, just with small subset actually: 

https://docs.kiwi.com/locations/#header-locale-setting

But it seems to be somewhat clever, e.g. https://api.skypicker.com/locations?term=PRG&locale=ar-MA `ar-MA` is not supported, but fallbacks to `ar-AE` at least, instead of English

jaroslav-kubicek: damn my response is related to locations 🙃

jaroslav-kubicek: this LocaleMap was prepared by @jakubzaba  and AFAIK it corresponds to what we have on kiwi

PR #85, opened 2018-04-23
Added FAQArticle voting

### LINKED ISSUE #69: Add FAQArticle voting mutation
Wrap this endpoint here https://gitlab.skypicker.com/django-team/knowledgebase#article-voting
to be able to vote or unvote articles of the FAQ.

Definition:

`mutation FAQArticleVote(vote: "up" | "down") : FAQArticle`

git-toni:
Closes #69


src/FAQ/mutations/FAQArticleVote.js
```diff
@@ -0,0 +1,69 @@
...
+  type: FAQArticle,
+  args: {
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'ID of FAQ article to receive vote.',
+    },
+    language: {
+      type: LanguageInput,
+      description: 'Language in which the article is returned.',
+    },
```
mrtnzlml: We are trying to move away from this approach. See: https://github.com/kiwicom/graphql/pull/82

src/FAQ/mutations/FAQArticleVote.js
```diff
@@ -0,0 +1,69 @@
...
+    vote: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'Is the vote positive or negative? ("up"/"down")',
```
mrtnzlml: Would enum be better maybe?

src/FAQ/mutations/FAQArticleVote.js
```diff
@@ -0,0 +1,69 @@
...
+
+import { post } from '../../common/services/HttpRequest';
+import config from '../../../config/application';
+import type { FAQArticleDetail as FAQArticleType } from '../dataloaders/getFAQArticle';
+import FAQArticle from '../types/outputs/FAQArticle';
+import { ProxiedError } from '../../common/services/errors/ProxiedError';
+import type { GraphqlContextType } from '../../common/services/GraphqlContext';
+import LanguageInput from '../../common/types/inputs/LanguageInput';
+
+const successfulResponse = { message: 'yummy' };
```
mrtnzlml: 🤔 Is this srsly what is the API returning? 🙈

src/FAQ/mutations/__tests__/FAQArticleVote.test.js
```diff
@@ -0,0 +1,69 @@
...
+describe('FAQArticleVote mutation', () => {
+  it('should vote UP', async () => {
+    RestApiMock.onPost(
+      config.restApiEndpoint.FAQArticleVote(goodId),
+    ).replyWithData({
+      message: 'yummy',
+    });
+    const voteMutation = `
+      mutation {
+        FAQArticleVote(id:"${globalId}", vote:"up") {
```
mrtnzlml: Can you please use second (I think) parameter of `graphql( ... )` function? The goal is to show best practices even in our tests...

src/FAQ/mutations/FAQArticleVote.js
git-toni: https://gitlab.skypicker.com/django-team/knowledgebase#article-voting :man_shrugging:

mrtnzlml: I mean... This is just... Embarrassing... 🙃

src/FAQ/mutations/FAQArticleVote.js
jaroslav-kubicek: Does this mean I should deprecate `Language` as well? 🙂

git-toni: I should probably leave this as it is for this PR? Seeing at how `getFAQArticle.js` currently works?

mrtnzlml: @jaroslav-kubicek What is the difference between `Locale` and `Language`? Seems quite similar but inconsistent.

jaroslav-kubicek:
> I identify user by token generated by combination of user-agent, ip and selected language

> So in most cases you can vote for single Article just once. Anyway you could later update your vote.

Can you investigate this? Because if there is only one instance of GraphQL, then I guess it would probably accept just single vote

jaroslav-kubicek:
I would probably suggest further update of rest api to use some sort of token to identify unique users

jaroslav-kubicek:
> He said textually: "user is identified by ip, user agent and selected language version; if user agent is missing, empty string is used".

Ok, then I guess we need to send unique `User-Agent` on every request because language + IP will be always same for multiple requests 🙂

git-toni:
Oh true, IP is the same as well, I was thinking of the user browser not the graphql server :man_facepalming:

git-toni:
So is some sort hashed `language+timestamp` combo good for you for this PR?

jaroslav-kubicek:
> So is some sort of hashed language+timestamp combo good for you for this PR? Or will backend complain we are "hacking" the vote limit?

Yes, just hash timestamp. I would keep it simple for now, later we can hash e.g. incoming `User-Agent` and `IP` from GraphQL client, but it's not necessary for now IMHO.

mrtnzlml:
How is random value going to prevent me from voting multiple times? Isn't this the main point why is it implemented on the server like this?

git-toni:
@mrtnzlml Yep, agree. That's what I meant by "hacking" the vote limit above. :smile:  
I guess we can choose between:
- Only one vote per article ever.
- Bypass the vote limit.
- Ask backend for some sort of whitelist of our `user-agent: graphql`. Ideal option, tho we should send BE some data so they re-implement the limit somehow.

git-toni:
Ok so Martin and Michal just let us know that this Voting endpoint is going to closed alltogether :sweat_smile: , so this PR makes no sense anymore. We will instead use article feedback/comments https://gitlab.skypicker.com/django-team/knowledgebase#article-comment

PR #92, opened 2018-04-26
Return city name when quering all available hotels with cityId

tbergq:


tbergq:
@mrtnzlml please review

tbergq:
Is there a way to use either or? in case it is, maybe we should integrate that? It doesn't really make sense to send both. 

say you send coordinats for barcelona and cityid for prague, what should happen?

mrtnzlml:
You actually described my concern. I was thinking about throwing an error in this case instead of allowing it because it really doesn't make any sense.

tbergq:
I created an issue

#110

PR #94, opened 2018-04-26
Add BookingInterface to differentiate between ONE_WAY, RETURN & MULTICITY

### LINKED ISSUE #76: Add "type" field to Booking type
Let's add `type` field to `Booking` type as this functionality is missing in current implementation and it's necessary to detect whether there is a return flight or multiple stopovers.

It should be an enum, currently with possible values `ONE_WAY`, `RETURN` and `MULTICITY`.
Definition of multicity - `segments` field in Booking API response is non-empty array AND there is no leg with `return=1` - needs further clarification that it's correct like this.

jaroslav-kubicek:
Closes #76 

(I don't like the implementation much, but it's not BC break & I would wait with a new query that works with interface until new API is prepared)


mrtnzlml:
@jaroslav-kubicek Tests are very unhappy. Can you please fix them before I actually check it?

jaroslav-kubicek:
I'm done.

mrtnzlml:
@jaroslav-kubicek Could you please also deprecate the rest of the fields in our `Booking` output type? We should use `onWay` field from now on to indicate that it's not that simple to fetch legs, right?

src/booking/datasets/booking-item-4903131.json
```diff
@@ -0,0 +1,345 @@
+{
```
mrtnzlml: @jaroslav-kubicek We are trying to reduce these snapshot files and not add them. Is it really necessary to add another? It's not even obvious why is this snapshot so special. I just assume it contains multi-city data? But only because you are just implementing it... :-/

jaroslav-kubicek: Is changing the filename enough? :pray: It was quite difficult to obtain some nice example of multicity booking

jaroslav-kubicek: As it’s the newest snapshot, I would like to use it to create flow type for booking api response and based on that implement one faker that will replace all these snapshots

mrtnzlml: Don't do it now, please. You are going to spend too much work on this. Just improve the snapshot name so I can merge it. These tests are not good anyway...

jaroslav-kubicek: fixed

Issue #63, opened 2018-04-18
Add CONTRIBUTING.md


oreqizer:
So I know how to do things

mrtnzlml:
@oreqizer What exactly is not clear so we'll make this file really helpful?

oreqizer:
## Things I missed so far

* API calls guide (using `HttpRequest`)
* How to test things
* More detailed folder structure
* File nomenclature (and nomenclature in general)
* How to update schema
* General info about CI, releasing etc.

PR #103, opened 2018-04-30
Currencies: Write tests

oreqizer:


src/currency/queries/__tests__/Currencies.spec.js
```diff
@@ -0,0 +1,34 @@
...
+import { graphql, RestApiMock } from '../../../common/services/TestingTools';
+import currenciesDataset from '../../datasets/currencies.json';
+
+beforeEach(() => {
+  RestApiMock.onGet(
+    'https://nitro-hankey.skypicker.com/currencies',
+  ).replyWithData(currenciesDataset);
+});
+
+describe('#Currencies', () => {
```
michalsanger: Why the `#`?

src/currency/queries/__tests__/Currency.spec.js
```diff
@@ -0,0 +1,48 @@
...
+import { graphql, RestApiMock } from '../../../common/services/TestingTools';
+import currenciesDataset from '../../datasets/currencies.json';
+
+beforeEach(() => {
+  RestApiMock.onGet(
+    'https://nitro-hankey.skypicker.com/currencies',
+  ).replyWithData(currenciesDataset);
+});
+
+describe('#Currencies', () => {
```
michalsanger: `Currency`

src/currency/queries/__tests__/__snapshots__/Currency.spec.js.snap
```diff
@@ -0,0 +1,27 @@
...
+      "round": 0,
+      "uncertainFormat": false,
+    },
+  },
+}
+`;
+
+exports[`#Currencies getting a non-existent currency 1`] = `
+Object {
+  "data": Object {
```
michalsanger: Some error message should be added. Try this query:
```
{
  location(id: "not_exist") {
    locationId
    name
    slug
  }
}
```

src/currency/queries/__tests__/Currencies.spec.js
oreqizer: looks nice in the output

michalsanger: But inconsistent :-(

oreqizer:
@michalsanger fixed all things

src/currency/queries/__tests__/Currencies.spec.js
oreqizer: k

mrtnzlml: > But inconsistent :-(

The same for `*.spec.js` for tests and so on. It's not the first inconsistency... :)

michalsanger: @mrtnzlml right, that can be fixed easily. Could you create an issue with other inconsistencies you know about? 

Looks like #63 is really needed

mrtnzlml: I don't think it makes sense without eslint rules or something like that. I miss a lot of things even with my OCD... 🙂 I knew about this even in the first PR but we allow it so it's not necessarily wrong. The same thing with the tests style. We usually use the BDD syntax `it` but not here. And again - it's not wrong, or is it?

I will be working on #63 today.

michalsanger: ESLint could be part of the PR. There is nothing like good/bad with this, it's only about agreement and write it up into `CONTRIBUTING.md`

PR #109, opened 2018-05-02
Tests: remove coverage reporting and coverage thresholds

mrtnzlml:
It leads to writing stupid snapshot tests or tests just to be OK with the coverage thresholds. I believe that better technique is to write tests when it actually makes sense (units for non-trivial logic and integrations for business critical things).

### DIFF
--- a/jest.json
+++ b/jest.json
@@ -4,17 +4,6 @@
     "src/**.js"
   ],
   "coverageReporters": [
-    "lcov",
-    "text",
     "text-summary"
-  ],
-  "coverageThreshold": {
-    "global": {
-      "branches": 80,
-      "functions": 80,
-      "lines": 80,
-      "statements": 80
-    }
-  },
-  "verbose": false
+  ]
 }


mrtnzlml:
Motivation: https://github.com/kiwicom/graphql/pull/94/files

Not a single test testing business logic (and there is a lot of business logic). Just _"here is the input, snapshot output and done"_... We should learn how to write meaningful tests instead.

tbergq:
I do agree, but I am not sure removing coverage report will improve testing by itself. 

We need to also start to write better tests.

mrtnzlml:
My internal motivation is to hide it. Write tests, sure, but not because of this number. That's the only reason why I sent this PR. Coverage will not decrease suddenly after this PR but the reasons why are we writing tests should change...

tbergq:
@michalsanger are you ok with removing this threshold?

michalsanger:
Why don't we just remove the `coverageThreshold` but keep the coverage info?

tbergq:
Is that ok @mrtnzlml? Will you make the change?

mrtnzlml:
It's faster without coverage reporting... 😀

mrtnzlml:
@tbergq Updated - what about this?

PR #111, opened 2018-05-02
Upgrade dependencies with only patch version changes

mrtnzlml:
No BC break expected.


mrtnzlml:
Tests are failing - no idea why... 😢

Issue #117, opened 2018-05-03
Allow querying allBookings from localhost


tbergq:
In production I can log in and execute query. On localhost I have to use postman, because there is no way to log in, or add access token.

![skjermbilde 2018-05-03 kl 12 08 23](https://user-images.githubusercontent.com/7705041/39571346-6fb53790-4ecb-11e8-9578-0c8a27d6e262.png)

jaroslav-kubicek:
Probably solved by #125

mrtnzlml:
I think the problem here is different. It's just hard and very unclear how to query bookings in our environment...

michalsanger:
Is the solution to opensource this GraphiQL? https://kiwi-graphiql.now.sh/

PR #118, opened 2018-05-03
Add hint to use GraphiQL app during development

jaroslav-kubicek:
Potentially useful info for newcomers to the project.


mrtnzlml:
Related issue: https://github.com/kiwicom/graphql/issues/117 (for this we need to use GraphiQL from our private repo 😞)

jaroslav-kubicek:
Desktop app has the ability to modify the HTTP headers, that was what triggered me initially to add this into readme 😎

mrtnzlml:
> Desktop app has the ability to modify the HTTP headers, that was what triggered me initially to add this into readme 😎

So you need to use Postman to get the header value? Or what do you mean? We need to login first using username and password...

jaroslav-kubicek:
you can execute mutation `login` to retrieve token

michalsanger:
@mrtnzlml this graphiql app supports setting headers https://github.com/skevy/graphiql-app

mrtnzlml:
> you can execute mutation login to retrieve token

Could you please add info on how to actually use it? If I remember correctly it's not that simple because you have to prefix the header value with `Token: ...` for example. 🙂 

Note: There is `DEV_API_TOKEN` in envs on BLKN to simplify this but nobody is probably using it...

PR #128, opened 2018-05-04
Add opaque ID

radkaag:


src/location/queries/Location.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 
 import { GraphQLString, GraphQLNonNull } from 'graphql';
+import { fromGlobalId } from 'graphql-relay';
```
mrtnzlml: @radkaag Please fix the failing CI check. This is wrong... :)

PR #137, opened 2018-05-08
Booking: un-deprecate some fields

mrtnzlml:
Basically allow all the information not related to the booking type to be fetched on the root Booking level as well. This will highly simplify the clients. You always have to use oneWay/return/multicity in order to fetch legs but that's not necessary for assets for example.

Cc @jaroslav-kubicek - I hope it makes sense (or we can deprecate it again 🙂)


jaroslav-kubicek:
aye, makes total sense 👍

PR #140, opened 2018-05-09
Booking: move 'destinationImageUrl' to the shared fields

mrtnzlml:
This will make destination image available everywhere (in bookings). I also fixed bug with default value of 'destinationImageUrl' field. It was necessary to refactor the helpers because of circular dependencies.


tbergq:
@mrtnzlml some test is unhappy

tbergq:
Then I am also happy 😃

PR #145, opened 2018-05-11
Add "carriers" field to Booking

jaroslav-kubicek:
I propose the idea to add this field, so we can easily retrieve a unique list of involved carriers and e.g. display a proper logo without the necessity to iterate over all `Trip.leg` and filtering on client.

I also updated the `airlines.json` dataset so it's up to date (Busses and trains were missing).

Controversion - should it be named as "carriers" (airlines+buses+trains) or stick with good old airlines?


mrtnzlml:
> Controversion - should it be named as "carriers" (airlines+buses+trains) or stick with good old airlines?

I think "carriers" is a great name.

src/booking/types/outputs/BookingInterface.js
```diff
@@ -94,6 +97,32 @@ export const commonFields = {
     resolve: BookingDestinationImageURL,
   },
 
+  carriers: {
+    type: new GraphQLList(Airline),
```
mrtnzlml: We should probably also focus on this since `Airline` is not accurate as you pointed out.

mrtnzlml: I am just not sure if it's a good idea to create a new field (`carriers`) with a type that is already virtually deprecated (`Airline`). It's going to be BC break as well to change it later so this field is already doomed and you should think about the future how are you going to replace it. My only advice is: do not kill your nice `carriers` before you even start... :)

jaroslav-kubicek: updated, it's still similar to `Airline`, as there is not much to add
- id field added
- logoUrl removed (displaying logo works differently now)

michalsanger:
@jaroslav-kubicek fix the conflict plz

jaroslav-kubicek:
@michalsanger  updated

src/booking/types/outputs/Carrier.js
```diff
@@ -0,0 +1,54 @@
...
+      description: 'Indicates whether it is low cost carrier.',
+      resolve: ({ isLowCost }: CarrierData): boolean => isLowCost,
+    },
+  },
+});
+
+export const getUniqueCarriers = (
+  carriers: $ReadOnlyArray<?Airline | ?CarrierData>,
+): CarrierData[] =>
+  carriers.reduce((acc, carrier) => {
```
michalsanger: When did you fell in love with `reduce` please? :-)

Can't we just use `uniqBy` from Rambda or Lodash? Or use `Set` as registry of know codes and simply iterate the `carriers` array?

src/flight/dataloaders/Airline.js
```diff
@@ -37,7 +37,7 @@ async function fetchAirlines() {
       // eslint-disable-next-line promise/prefer-await-to-then
       .then(airlines => {
         const cache: { [key: string]: Airline } = {};
-        airlines.map((a: Airline) => (cache[a.id] = a));
+        airlines.forEach((a: Airline) => (cache[a.id] = a));
```
michalsanger: Is there any difference?

jaroslav-kubicek: sry, I couldn't resist :) 

map => getting new array from the old one
forEach => doing some side effects for each item in the array

src/booking/types/outputs/Carrier.js
jaroslav-kubicek: When I heard about the functional approach in programming for the first time.

reduce is normal, quite convenient function, no need to have dozens of immediate variables to do the same thing.

src/flight/dataloaders/Airline.js
jaroslav-kubicek: it could be refactored to reduce as well btw... 👼

PR #147, opened 2018-05-14
Add booked services field to bookings query

tbergq:


src/booking/types/outputs/BookedService.js
```diff
@@ -0,0 +1,24 @@
...
+export default new GraphQLObjectType({
+  name: 'BookedService',
+  fields: {
+    category: {
+      type: GraphQLString,
+      description: 'Type of booked service',
+      resolve: ({ category }: BookedServiceType) => category,
+    },
+    status: {
+      type: GraphQLString,
```
mrtnzlml: Can we use enum here? It simplifies implementation of Relay clients (typed clients in general).

tbergq: Sure, need to know the statuses, I will see what I can find out

src/booking/types/outputs/BookingInterface.js
```diff
@@ -6,19 +6,26 @@ import {
...
 import GraphQLBookingStatus from '../enums/BookingStatus';
 import Price from '../../../common/types/outputs/Price';
 import BookingDestinationImageURL from '../../resolvers/BookingDestinationImageURL';
+import GraphQLBookedServices from './BookedService';
 
 export type BookingInterfaceData = BookingsItem;
 
+const { connectionType: BookedServicesConnection } = connectionDefinitions({
+  nodeType: GraphQLBookedServices,
+});
```
mrtnzlml: Do we really need the connection here? What about a simple list? I don't think we are ever going to paginate our services.

src/booking/types/outputs/BookedService.js
mrtnzlml: Don't worry. I believe enum can be extended without BC breaks and Flow actually forces you to write your clients to be compatible with this change.

src/booking/types/outputs/BookedService.js
```diff
@@ -0,0 +1,24 @@
...
+
+type BookedServiceType = {|
+  category: string,
+  status: string,
+|};
+
+export default new GraphQLObjectType({
+  name: 'BookedService',
+  fields: {
+    category: {
```
mrtnzlml: "Category" also sounds enum-ish. What is it actually? Do you have an example of possible values?

tbergq: I wish I did, but I will need it sooner or later, so I might as well find out now

tbergq:
@mrtnzlml please check again

PR #148, opened 2018-05-14
Added guarantee field to Leg that return an new enum type.

edgarordonez:
Close #146 

I added guarantee field to Leg with an enum to return KIWICOM or CARRIER guarantee.


michalsanger:
Would be cool to add the new property into query in some test to see it really works.

michalsanger:
Can you please rebase this PR on master? Looks like there is also some feature branch merged into this one.

PR #152, opened 2018-05-16
Add passenger data to booking interface

tbergq:


src/booking/types/outputs/Passenger.js
```diff
@@ -0,0 +1,62 @@
...
+
+const InsuranceTypeEnum = new GraphQLEnumType({
+  name: 'InsuranceType',
+  description: 'The possible insurance type values',
+  values: {
+    NONE: { value: 'none' },
+    TRAVEL_BASIC: { value: 'travel_basic' },
+    TRAVEL_PLUS: { value: 'travel_plus' },
+    SKYSILVER: { value: 'skysilver' }, // Deprecated, same as travel_basic
+    SKYGOLD: { value: 'skygold' }, // Deprecated, same as travle_plus
```
mrtnzlml: Do not expose it, please. Hide it behind `TRAVEL_BASIC` and `TRAVEL_PLUS`.

tbergq: How, an enum is a key value map?

mrtnzlml: You have to tweak the values in the resolver. We surely don't want to expose it since it's already deprecated and mostly removed.

tbergq: Right, that should work, thanx

tbergq: @mrtnzlml please check again

PR #153, opened 2018-05-16
Revert "Set booking interface global id to type Booking"

tbergq:
This reverts commit 9cbf514ccd9fd893eec3eadabdbbb68d947e98dd.


mrtnzlml:
It's actually BC break so let's hope no one actually implemented this in last 24 hours... 😊

tbergq:
> Congratulations! NO BREAKING CHANGES or OUTDATED SCHEMA. Good job!

We should update test then 🙈

mrtnzlml:
This test checks something different. We didn't have a test for this at all... :)

PR #154, opened 2018-05-17
Add passenger full name field

tbergq:


michalsanger:
No tests? :-(

We did it little bit more robust here:
https://github.com/kiwicom/graphql/blob/e51d03d37198ae85bc2edf98d0397d79e99d59aa/src/identity/types/outputs/Identity.js#L42-L52

mrtnzlml:
What do you want to test here? However, it's true that the `Identity` implementation is way more robust so maybe can you please reuse this logic @tbergq?

michalsanger:
As I see on the web, both fields are required, but it's simple one line to handle edge cases so why not..

I'd add the field into some passengers query and be happy it's resolved :-)

tbergq:
Sure, I am on it, also just to point out, your implementation does not handel firstname or lastname undefined

mrtnzlml:
@tbergq Please fix this simple one line in `Identity` as well... :)

PR #155, opened 2018-05-17
Improve booking fullname, add test, improve logic in Identity fullname

tbergq:


src/booking/queries/__tests__/Booking.test.js
```diff
@@ -124,4 +124,79 @@ describe('single booking query', () => {
     }`;
     expect(await graphql(query)).toMatchSnapshot();
   });
+
+  it('creates fullname correctly', async () => {
```
mrtnzlml: Personally, I would prefer to (unit) test the resolver directly because this is what you actually want to test (at least according to the test title). Do you want to improve it or nah? I am OK to merge even this version... :)

src/identity/types/outputs/Identity.js
```diff
@@ -44,10 +44,10 @@ export default new GraphQLObjectType({
       description:
         'Concatenation of first and last name with fallback to the login field.',
       resolve: ({ firstName, lastName, login }: Identity): string => {
-        if (firstName === null && lastName === null) {
+        if (firstName == null && lastName == null) {
           return login;
         }
-        return [firstName, lastName].filter(word => word !== null).join(' ');
+        return [firstName, lastName].filter(word => word != null).join(' ');
```
mrtnzlml: Please unit test this as well because you are fixing it.

tbergq: @mrtnzlml please check again

src/booking/queries/__tests__/Booking.test.js
```diff
@@ -124,4 +124,79 @@ describe('single booking query', () => {
...
+    RestApiMock.onGet(
+      'https://booking-api.skypicker.com/api/v0.1/users/self/bookings/2707251?simple_token=b206db64-718f-4608-babb-0b8abe6e1b9d',
+    ).replyWithData({
+      ...SingleBooking,
+      passengers: [
+        {
+          category: 'adults',
```
michalsanger: Are all the passenger props needed since the query is using only `fullName`? Would be clear what's the point of the test without them I'd say.

tbergq: I will remove it since I added the unit test for this anyway

mrtnzlml: @michalsanger I believe sanitizers would not be happy at some point?

mrtnzlml: @tbergq You can test just the query validity if you want ([example](https://github.com/kiwicom/graphql/blob/971dd0f37d32b426759bfc16208bd7cbca62e04d/src/booking/queries/__tests__/AllBookings.only.test.js)). Just to make sure the query (field) will work forever... :)

PR #163, opened 2018-05-22
Booking: Add visaInformation to passenger

### LINKED ISSUE #157: Bookings: Add visa information

tbergq:
closes #157


src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -134,9 +134,25 @@ function sanitizePassengers(passengers: Object[]) {
...
   }));
 }
 
+function sanitizeVisaStatus(travelInfo: Object, passengerId: number) {
+  const outData = [];
+
+  // TravelInfo object contains keys which are passenger ids
+  // Return only data for the current passenger
+  if (travelInfo != null && travelInfo.hasOwnProperty(passengerId.toString())) {
+    const passengerTravelInfo = travelInfo[`${passengerId}`];
```
mrtnzlml: Is `` `${passengerId}` `` necessary here? Looks funny... 🙂

src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
```diff
@@ -143,6 +143,34 @@ Object {
         "expiration": 1608768000,
         "idNumber": "12345678XY",
       },
+      "travelInfo": Array [
+        Object {
+          "visa": Array [
+            Object {
+              "country": "PL",
```
mrtnzlml: Why is here only the code if the output type is `Location`?

mrtnzlml: Oh I see - it's sanitizer...

src/booking/dataloaders/ApiSanitizer.js
tbergq: You are right, it is not necessary, updated

scripts/graphql-schema-snapshot.json
```diff
@@ -2607,6 +2619,65 @@
...
+                "name": "Location",
+                "ofType": null
+              }
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
+            "name": "notice",
+            "description": "Countries with visa status critical",
```
mrtnzlml: Wrong description (it's a notice, not critical).

src/booking/types/outputs/Visa.js
```diff
@@ -0,0 +1,51 @@
...
+
+export default new GraphQLObjectType({
+  name: 'Visa',
+  description: 'Visa information for passenger',
+  fields: {
+    critical: {
+      type: new GraphQLList(GraphQLLocation),
+      description: 'Countries with visa status critical',
+      resolve: async (
+        { critical }: { critical: string[] },
```
mrtnzlml: Can these types be exact?

src/booking/types/outputs/Visa.js
```diff
@@ -0,0 +1,51 @@
+// @flow
+
+import { GraphQLObjectType, GraphQLList } from 'graphql';
+
+import GraphQLLocation from '../../../location/types/outputs/Location';
+import type { GraphqlContextType } from '../../../common/services/GraphqlContext';
+
+const loadLocations = async (dataLoader: Object, countryCodes: string[]) => {
+  const promises = countryCodes.map(code => dataLoader.location.load(code));
```
mrtnzlml: There is no `loadMany` anymore? I know there were changes around location loaders but we should really use this method from the data-loader...

tbergq:
Updated, please check again @mrtnzlml

src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
```diff
@@ -143,6 +143,34 @@ Object {
...
+The max. stay is granted within 180 days.
+</p>
+
+</div>",
+              "status": "critical",
+              "timestamp": 1447264497,
+            },
+            Object {
+              "country": "PL",
+              "info": "<div class=\\"content\\"><p><b>The passenger is arriving on a domestic or common border area flight and no immigration regulations apply for this segment.  See Entry Requirements for previous segment.</b></p></div>",
```
mrtnzlml: Are we going to be able to fetch these lines in a raw format if needed? Also, how would we add them to our `Visa` type? I am just thinking about the future evolution of this type - are `ok`, `notice` and `critical` the best names of these fields? It's more about the countries...

tbergq: I am open to suggestions for a better design, it is very confusing to work with the existing api 😖

mrtnzlml: What about this:

```
visa
 - requiredIn (cities, locations)
 - warningIn
 - optionalIn
```

I really don't know. It's not much better than the `ok`, `notice` and `critical`. Maybe even worse... :) Let me ask on Slack.

tbergq: Not sure about these names, I don't think optional is good naming. It is not optional, it is not needed.

Not sure about warning either, it is not a warning, just a notice to check it your self. Though that could also be considered a warning 🤔  
requiredIn sound like good naming in that case.

tbergq:
@mrtnzlml changed names to
- requiredIn
- warningIn
- okIn

what do you think?

mrtnzlml:
What about `noticeIn` as you suggested? Perhaps we don't need the `okIn` at all? Because we have to check whether it's OK and if not then we need to know where is it wrong...

tbergq:
I am ok with `warningIn` also, it gives you a warning so 🤷‍♂️ 
okIn is probably not needed in the client, since we show ok if there is no warning nor required, but still, I think it makes sense to do this on the `graphql`, atleast just to verify that it works.

PR #164, opened 2018-05-22
Fix error in visaInformation

tbergq:


src/booking/types/outputs/Visa.js
```diff
@@ -13,33 +19,33 @@ export default new GraphQLObjectType({
       type: new GraphQLList(GraphQLLocation),
       description: 'Countries where visa is required',
       resolve: async (
-        { critical }: {| critical: string[] |},
```
tbergq: Seems this typing is far from optimal 🙈

PR #165, opened 2018-05-28
GraphQL: add new 'availableWhitelabeledServices' query

mrtnzlml:


src/tripServices/dataloaders/Lounges.js
```diff
@@ -0,0 +1,39 @@
...
+import { fetchJson } from '../../common/services/JsonFetcher';
+
+type LoungeInfoType = {|
+  +airport: string,
+  +terminal: string,
+  +iata: string,
+  +lounge_name: string,
+  +lounge_code: string,
+  +provider: 'collinsons' | 'loungebuddy' | '%future added value',
+|};
```
tbergq: `'%future added value'` should this rather be a comment?

mrtnzlml: definitely not - it must be part of the code

src/RootQuery.js
```diff
@@ -47,5 +48,6 @@ export default new GraphQLObjectType({
     allFAQCategories: AllFAQCategories,
     FAQCategory: FAQCategory,
     FAQArticle: FAQArticle,
+    availableTripServices: AvailableTripServices,
```
michalsanger: Could we name it `allTripServices` for consistency?

src/tripServices/dataloaders/Lounges.js
michalsanger: What's the purpose? Placeholder? Comment in code would be great.

src/tripServices/dataloaders/Lounges.js
```diff
@@ -0,0 +1,39 @@
...
+
+export default new DataLoader(async (keys: $ReadOnlyArray<string>): Promise<
+  $ReadOnlyArray<LoungeInfoType[]>,
+> => {
+  const stringifiedKeys = keys.join(',');
+
+  // [
```
michalsanger: Is this comment needed?

src/tripServices/dataloaders/Lounges.js
```diff
@@ -0,0 +1,39 @@
...
+
+import DataLoader from 'dataloader';
+
+import { fetchJson } from '../../common/services/JsonFetcher';
+
+type LoungeInfoType = {|
+  +airport: string,
+  +terminal: string,
+  +iata: string,
+  +lounge_name: string,
```
michalsanger: Props should be in camelCase

src/RootQuery.js
mrtnzlml: ❌ I don't want to name it like this just because of consistency. It doesn't make any sense.

michalsanger: It's not naming convention to prefix with `all` queries returning list of items?

tbergq: I thought our `all...` queries where returning some connection. 

This query is not returning a connection.

mrtnzlml: > This query is not returning a connection.

More importantly, this query is not going to return ALL trip services but only the available ones.

src/tripServices/dataloaders/Lounges.js
mrtnzlml: Why is this so strange for you both? 😄 It's for _"future added value"_. The same thing is used in Relay - you never noticed? I am not sure if I want to comment every single enum in Flow...

src/tripServices/dataloaders/Lounges.js
mrtnzlml: Yes, I think so - this is why I added it... 🙂

src/tripServices/dataloaders/Lounges.js
tbergq: No, I have not noticed. I just don't se why it is better than just

`+provider: 'collinsons' | 'loungebuddy'`

mrtnzlml: Because it's not enough for this kind of enum. I will remove it... 🙄

tbergq: https://github.com/facebook/relay/issues/2351

tbergq: > I will remove it.

I am ok to keep it

mrtnzlml: @tbergq It's everywhere in our mobile codebase (`__generated__`) because we use latest Relay. Try to search for _"%future added value"_ string. You will be maybe surprised... 🙂

src/RootQuery.js
michalsanger: Does it have to be a query anyway? Doesn't make more sense to have `Trip(iataCode: String!)` query and `availableServices` as a `Trip` entity property?

src/tripServices/dataloaders/Lounges.js
michalsanger: I'm OK with keeping it, just as we can see at least 2 devs were surprised, so please add the comment for next readers.

mrtnzlml: I removed it so we will have no more confused developers.

src/RootQuery.js
mrtnzlml: I am just interested in what services we provide for this one IATA code. That's it - no trips, not cities, nothing more. I don't see your point very clearly. Can you please expand more on your idea?

michalsanger: Just feel little bit strange with providing specific queries like this. If it's connected with some entity I'd rather return the data via property of that entity. For example we probably wouldn't have query `allTripLegs`.

Also back to the naming, how can a Trip be identified by IATA code? That code is for airports and airlines.

mrtnzlml: > For example we probably wouldn't have query `allTripLegs`.

Why not? I am OK with that if you need it.

> Also back to the naming, how can a Trip be identified by IATA code? That code is for airports and airlines.

I don't know. I am not fetching trips here. I am fetching all available services for any trip.

michalsanger: > Why not? I am OK with that if you need it.

Because it's universal just to have "entity query" `Trip`

```graphql
{Trip(id: "...") {legs {...}}}
{Trip(id: "...") {availableServices {...}}}
```

> I am fetching all available services for any trip

But so far trip is "Single travel from origin to destination, with possible stopovers." so hardly can be identified by one IATA code. Aren't you asking for service available on given airport?

mrtnzlml: yes

michalsanger: So it's not related to Trip but Location...

mrtnzlml: Can you please propose the query name?

mrtnzlml: @michalsanger Would you be happy with just `availableServices`?

mrtnzlml: `allAvailableServices` 🙄

michalsanger: I'd rather go with the "entity property" proposal. 
```graphql
Location(iataCode: String!) {avalableServices {...}}
```

mrtnzlml: I would not... 🙂

@tbergq please decide and I will make it

tbergq: What is the use case here, what are the arguments for keeping it as a query vs the arguments for not having it as a part of the location query?

mrtnzlml: > What is the use case here...

I need it to render "Trip Services" screen in the mobile app. You'll get IATA and you need to know what services are available.

> ...what are the arguments for keeping it as a query...

It's exactly what I need for my use-case at this moment. I can easily move it to the field under `Location` query if my requirements change.

> ...the arguments for not having it as a part of the location query?

`Location` query currently accepts everything as an input parameter. So the proposal is to create new input parameter `iataCode`. It's weird in both cases:

- field `availableServices` won't work with `id` input or only sometimes (when it's actually IATA code), or
- you have to use `iataCode` input in order to fetch this one field

Solution is to describe it in the description and make it even more complicated and not very well readable? I propose to solve it by design instead - `availableServices` query accepts only IATA without quirks. Moreover, I am really not interested in the location so I don't see a point in calling location API (this may change in the future).

tbergq: Then I give my vote to keep it as a separate query

mrtnzlml: Let's wait for @michalsanger...

michalsanger: The "Trip Services" screen does not display location name?

> field availableServices won't work with id input or only sometimes (when it's actually IATA code)

The "won't work" means "it will be empty array" that's OK IMHO.

Fetching location by IATA code sounds like OK requirement to me.

Anyway I'm OK with `availableServices` query if we in rush here

src/tripServices/types/output/TripServices.js
```diff
@@ -0,0 +1,30 @@
...
+import { GraphQLDateTime } from 'graphql-iso-date';
+
+import LoungeWhiteLabelURLResolver from '../../resolvers/LoungeWhitelabelURL';
+
+type AncestorType = {|
+  +iataCode: string,
+|};
+
+export default new GraphQLObjectType({
+  name: 'TripServices',
```
mrtnzlml: FIXME: rename and do not return `loungeWhitelabelURL` directly - we have to make sure that this service is actually available first (must be always part of the response), something like this could work:

```graphql
type AvailableServices {
  lounge: Lounge
  carRentals: CarRentals
  ...
}

type Lounge {
  whitelabelURL: string
}
```

src/RootQuery.js
mrtnzlml: > The "Trip Services" screen does not display location name?

No. There are locations deeper in the tree but we don't fetch them currently at once. I can connect locations to this query as well... 🙂 

> The "won't work" means "it will be empty array" that's OK IMHO.

There is no array. It will return for example `"lounge": null`. I should change the graph though... (https://github.com/kiwicom/graphql/pull/165/files#r191231666)

mrtnzlml: Currently max one lounge service and one car rental service on the airport yes. It's actually not exactly true because we return different whitelabel URLs depending on the conditions but from the client perspective, it's only one service.

I could the same way add it to the "booking" queries. It would make even more sense than in locations. But I have currently different needs on my client...

michalsanger: I was expecting array of services, because that matches the reality. Honestly this really looks like temp query.

mrtnzlml: Let's consider it to be a temporary query then... 😄

mrtnzlml:
@tbergq Please check again. I changed the query a bit:

```
{
  availableServices(iataCode: "PRG") {
    lounge(departureTime: "2007-12-03T10:15:30Z") {
      whitelabelURL
    }
  }
}
```

It will return `null` if it's not available at all (and therefore the button is going to be hidden). It returns an object if everything is fine. @michalsanger also pointed out that we could (should?) return an array here. It makes sense. However, we have to keep strict order of the services:

1. ~**insurance (not in this API)**~ _(this service is very special and not part of the group)_
2. car rental (from this API)
3. **hotel (not in this API)**
4. lounge (from this API)
5. parking (from this API?)
6. transportation (from this API?)

I find it more complicated because I don't want to iterate the services on the client (we are slowly replacing them with the native implementation - see hotels). I would also have to expose the type because this API will not take care of the translations, icons and other additional design details.

My intention from the very beginning is **not to write a loop** in the client to iterate the services.

For this reason I designed it like this so we can access the services directly in the query. I would propose to create a new query in case we'll need this.

mrtnzlml:
Note: the reason why I named it _"trip services"_ is because we have **trip** and **flight** services and we should not confuse them together. There are not going to be things like seats, baggage or pets...

mrtnzlml:
@tbergq Renamed to the `availableWhitelabeledServices` query as we discussed. We can easily deprecate it in the future. I am done here, please check... 🙂

PR #166, opened 2018-05-28
Bookings: Add isPastBooking field

tbergq:
I know there is PAST | FUTURE field when querying allBookings, but this is useful when querying booking(id). 

There are different things to be done on client when a flight is past vs future. 
It makes sense to be able to access this property easily.


src/booking/queries/__tests__/AllBookings.isPastBooking.test.js
```diff
@@ -0,0 +1,36 @@
...
+    `${allBookings}/2707224\\?simple_token=[0-9a-f-]{36}`,
+  ).replyWithData(Booking2707224Dataset);
+});
+
+it('should return isPastBooking', async () => {
+  const query = `{
+      allBookings {
+        edges {
+          node {
+            isPastBooking
```
mrtnzlml: @tbergq Is your intention to test here the return value or are you just making sure the query is fine (will work in the future)?

tbergq: Making sure the query is fine.

There are other tests for the isBooking function.

mrtnzlml: You can just use `validate( ... )` to make sure the query is not invalid. But I will accept it.

tbergq: Ok, I will definitely use that next time 🙂

Issue #175, opened 2018-05-29
Evolve 'availableWhitelabeledServices' query


mrtnzlml:
Now we finally know what we actually need to build (hopefully) so after some discussion with @tbergq I am going to turn +- this query into the new field under the booking query. It will introduce BC break:

```
query TripServicesQuery($departureTime: DateTime!) {
  booking(id: $id) {
    availableWhitelabeledServices {
      lounge(departureTime: $departureTime) {
        whitelabelURL
        airports {
          whitelabelURL
          location: Location
        }
      }
      parking {
        whitelabelURL
      }
      carRental {
        whitelabelURL
        cities {
          whitelabelURL
          location: Location
        }
      }
    }
  }
}
```

Turned out that the logic when the services are or are not available is far more complicated and it's tightly coupled with booking ID (each service differently).

mrtnzlml:
@tbergq I maybe did a little mistake here again. I ended up with this query:

```graphql
query TripServicesQuery {
  booking(id: 1234567) {
    availableWhitelabeledServices {
      lounge(departureTime: "2007-12-03T10:15:30Z") {
        relevantAirports {
          whitelabelURL
          location {
            slug
          }
        }
      }
      parking(fromDate: "2007-12-03T10:15:30Z", toDate: "2007-12-24T10:15:30Z") {
        whitelabelURL
      }
    }
  }
}
```

I did it because I thought it's closely related to the booking. And it is. But I think this would be a better approach (the query is not under the booking):

```graphql
query TripServicesQuery {
  availableWhitelabeledServices(bookingId: 1234567) {
    lounge(departureTime: "2007-12-03T10:15:30Z") {
      relevantAirports {
        whitelabelURL
        location {
          slug
        }
      }
    }
    parking(fromDate: "2007-12-03T10:15:30Z", toDate: "2007-12-24T10:15:30Z") {
      whitelabelURL
    }
  }
}
```

I am not going to refactor it now because this API is already a little monster but I am interested in our opinion. I am not sure if it's a good idea to have one huge type (`Booking`). This is IMHO better and you can send two queries in one request if you need booking as well. Data loaders will handle it easily...

PR #177, opened 2018-05-30
Move 'availableWhitelabeledServices' query under booking query

mrtnzlml:
This will allow me to do the magic in GraphQL and not in the client.

Closes: https://github.com/kiwicom/graphql/issues/175


src/booking/types/outputs/services/LoungeService.js
```diff
@@ -0,0 +1,50 @@
...
+              }: {|
+                +iataCode: string,
+                +departureTime: Date,
+              |}) => {
+                return LoungeWhiteLabelURLResolver(iataCode, departureTime);
+              },
+            },
+            location: {
+              type: Location,
+              resolve: ({ iataCode }, _, context) => {
```
tbergq: missing flow types

mrtnzlml: Do you need them? 🙂

mrtnzlml: I will refactor the types here...

PR #193, opened 2018-06-08
Dependencies: Upgrade apollo-engine, express and idx

tbergq:


tbergq:
😞 the tests does work locally 🤷‍♂️

PR #194, opened 2018-06-08
Add new 'customerBookings' root query

mrtnzlml:
Introducing new shiny query. This query replaces the old `allBookings` query as well as `booking` query (use `node` instead).

BTW the way how we work with `node` is incredibly painful...


src/booking/queries/Booking.js
```diff
@@ -10,6 +10,7 @@ import type { BookingsItem } from '../Booking';
 export default {
   type: GraphQLBooking, // may be null (doesn't exist)
   description: 'Find booking by its id or databaseId.',
+  deprecationReason: 'Use "node" query instead.',
```
jaroslav-kubicek: I thought we don't like `node` query much 🙂

michalsanger:
Do we have any naming strategy? I thought that query returning connection should be prefixed with `all`

tbergq:
Why do we need to prefix all queries with `all`? is not the naming clarifying enough?

michalsanger:
I was used to this:
 - prefix with `all` if query returns connection
 - use plural if query returns list

I agree it's no big help, just wanna make it clear if this convention is useful for us.

tbergq:
I was taught that prefixing and suffixing is a bad naming practice. 

If every file is named like `MMBCityImage` and `MMBFlightList` and so on, it makes readability worse than just `CityImage` and `FlightList`

src/booking/queries/Booking.js
mrtnzlml: We (I) don't like `node` query implementation much. Otherwise, it's brilliant.

mrtnzlml:
No such a convention. Unless we have eslint rule for it (or something that will help with it automatically). It's also a little bit odd to ask for `allFlights(onlySomething: true)`...

michalsanger:
OK, should we deprecate those `all` queries?

mrtnzlml:
Just keep evolving. No need to spend hours on rewriting it now. Maybe we'll go for `allCustomerBookings` after we'll discover how I messed it up again.

michalsanger:
I didn't mean right now :-) I just prefer naming consistency you know...

mrtnzlml:
Yes, I know. However, there are big objections against this convention and I want to step out and try it differently. Let's see... :)

PR #195, opened 2018-06-08
Dependencies: upgrade express-graphql

tbergq:


mrtnzlml:
Seems like we have to upgrade Babel to transpile `**` for older JS...

PR #203, opened 2018-06-11
Added all valid booking types

dinodsaurus:
On smart-faq we use combination of `BookingInterfaceConnection` and `booking` so this fix is needed to support those changes. 
More info at https://github.com/kiwicom/smart-faq/pull/392


mrtnzlml:
@dinodsaurus You should use `node` query for this (this is what you are essentially doing here). This "opaque ID" branch in `booking` query is deprecated.

mrtnzlml:
Maybe please describe better your use-case before I close this. You should avoid this in all cases (there are ways how to do it)...

dinodsaurus:
hey @mrtnzlml sorry I rolled back `smart-faq` PR so you cannot see the conversation we had there..
I refactored everything to `node` but @jaroslav-kubicek pointed out a big issue.
The problem is that from `kiwi-frontend` we receive just an `id: number` so we need to use `booking` and not `node`. 😅

mrtnzlml:
@dinodsaurus I understand but this will not fix it because you hot-patched the `ID!` branch. Since you are using `databaseId` only with this query it won't be even executed (and that's why the `ID!` branch will be removed soon).

jaroslav-kubicek:
Unfortunately, we rely on the possibility to pass both `databaseId` and `ID`
- `ID` is passed when you select one specific booking from a list of all bookings
- `databaseId` is passed fro frontend in MMB page

if we want to switch from `allBoookings` to `customerBookings`, this is the only viable way for us at the moment, otherwise we will stick with `allBookings` forever.

mrtnzlml:
You should fetch both IDs then and use them in appropriate queries (or work only with `databaseId` if you have to). Again: this feature is **deprecated** so please do not try to build new software on top of it...

mrtnzlml:
Closing - all explained and will be fixed on the SmartFAQ client.

PR #207, opened 2018-06-11
Remove AWS Lambda specific code

mrtnzlml:
This also revealed we are not testing Docker builds! :scream:


mrtnzlml:
_source_

**Update on GraphQL from the DevOps point of view**

Current state is that we have several separated GraphQL endpoints (locations, ?***, now I see ancillaries). The plan is to unify it and have one endpoint. This is how GraphQL is designed. All client libraries and tools expect that.

- from traffic point of view we will have to run GraphQL "near" flights service

- how to run it: API Gateway/AppSync + Lambda vs Containerized. Serverless solution looks much more expansive (Michal Sänger wants to do some research about that) and also is not so well supported on GCP. So probably we will go container-way.

PR #208, opened 2018-06-13
Adding bookingTimeline query

RobinCsl:
Basic implementation of the bookingTimeline query.
The assumption is that we know what to do with field `data` given the type of the event.


src/booking/queries/BookingTimeline.js
```diff
@@ -0,0 +1,49 @@
...
+
+import generateEventsFrom from '../resolvers/BookingTimeline';
+
+export default {
+  type: GraphQLBookingTimeline,
+  description: 'Get the timeline for your booking.',
+  args: {
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'Both Booking id & databaseId can be used as ID.',
```
mrtnzlml: That's not true.

src/booking/types/outputs/BookingTimelineEvent.js
```diff
@@ -0,0 +1,40 @@
...
+  name: 'TimelineEvent',
+  fields: {
+    timestamp: {
+      type: GraphQLDateTime,
+    },
+    type: {
+      type: GraphQLEventTypeEnum,
+    },
+    data: {
+      type: GraphQLString,
```
mrtnzlml: Is this the best way how to do it? I have no idea how to use it, what does it mean, what can I expect here. That's not very good...

mrtnzlml: Try to image the client implementation at the same time. If I only knew this is the only filed we will ever need it would be cool. I am really afraid that's not the case. You don' have to rewrite it just because I said so - if you convince me this is the good way how to do it, I will be fine with that... :)

src/booking/queries/__tests__/BookingTimeline.test.js
```diff
@@ -0,0 +1,142 @@
...
+});
+
+describe('single booking timeline query', () => {
+  it('should be of BookingTimeline type', () => {
+    expect(BookingTimeline.type.toString()).toBe('BookingTimeline');
+  });
+
+  it('should not accept database id', async () => {
+    const databaseIdQuery = `{
+      bookingTimeline(id: 2707251) {
```
mrtnzlml: No need to use the whole query just to verify the parameter doesn't work. This would be enough:

```graphql
{
  bookingTimeline(id: 2707251) {
    __typename
  }
}
```

src/booking/queries/__tests__/BookingTimeline.test.js
```diff
@@ -0,0 +1,142 @@
...
+      }
+        }
+      }
+    }`;
+    expect(await graphql(databaseIdQuery)).toMatchSnapshot();
+  });
+
+  it('should work with opaque Booking id', async () => {
+    const query = `{
+      bookingTimeline(id: "Qm9va2luZzoyNzA3MjUx") {
```
mrtnzlml: I would move the `__typename` one level up so you don't have to repeat it everytime:

```graphql
{
  bookingTimeline(id: "Qm9va2luZzoyNzA3MjUx") {
    __typename
    ... on Something {
      timestamp
    }
    ... on SomethingElse {
      receiptUrl
    }
    # ...
  }
}
```

src/booking/types/outputs/BookingTimelineEvent.js
```diff
@@ -0,0 +1,168 @@
...
+  },
+  isTypeOf: value => value.type === 'DownloadETicket',
+});
+
+export const DownloadBoardingPass = new GraphQLObjectType({
+  name: 'DownloadBoardingPass',
+  fields: {
+    timestamp: {
+      type: GraphQLDateTime,
+    },
```
mrtnzlml: This even is just saying "hey, you should download boarding pass" without providing URL of the boarding pass? 😊

RobinCsl: For now, I could only see the link to the ETicket, so I just created the event not to forget about it, but there is no data for now.

Issue #212, opened 2018-06-13
Proxied errors are not compatible with the latest GraphQL specification


mrtnzlml:
Currently, we report this kind of proxied errors:

```json
{
  "data": {
    "resetPassword": null
  },
  "errors": [
    {
      "_proxy": {
        "statusCode": null,
        "url": "https://auth.skypicker.com/v1/user.resetPassword"
      },
      "locations": [
        {
          "column": 32,
          "line": 1
        }
      ],
      "message": "User is not verified.",
      "path": [
        "resetPassword"
      ]
    }
  ]
}
```

However, this is incompatible with the latest GraphQL specification. We should use `extensions` for this. See: http://facebook.github.io/graphql/draft/#sec-Errors

> Previous versions of this spec did not describe the extensions entry for error formatting. While non‐specified entries are not violations, they are still discouraged.

mrtnzlml:
Note: we need to a way how to migrate easily since they are using it on SmartFAQ. (do not delete the `_proxy` key right away)

mrtnzlml:
Awaiting: https://github.com/kiwicom/smart-faq/pull/446

PR #214, opened 2018-06-14
Add 'extensions' to the errors response

mrtnzlml:
Current error object looks like this:

```json
{
  "_proxy": {
    "statusCode": "123",
    "url": "http://a.b"
  },
  "extensions": {
    "proxy": {
      "statusCode": "123",
      "url": "http://a.b"
    }
  },
  "message": "custom message",
  "locations": [
    {
      "line": 4,
      "column": 5
    }
  ],
  "path": [
    "currency",
    "format"
  ]
}
```

It should be backward compatible. See: http://facebook.github.io/graphql/draft/#sec-Errors

Related issue: https://github.com/kiwicom/graphql/issues/212

### DIFF
--- a/src/common/services/errors/ProxiedError.js
+++ b/src/common/services/errors/ProxiedError.js
@@ -1,21 +1,37 @@
 // @flow
 
+type ProxiedErrorValues = {|
+  +statusCode: string,
+  +url: string,
+|};
+
 /**
  * This exception is thrown if origin server does NOT respond with HTTP status code 200.
  */
-function ProxiedError(
-  message: string,
-  originStatusCode: number | string,
-  originUrl: string,
-) {
-  this.name = 'ProxiedError';
-  this.message = message;
-  this.originStatusCode = Number(originStatusCode);
-  this.originUrl = originUrl;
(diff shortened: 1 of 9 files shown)


src/common/services/errors/ProxiedError.js
```diff
@@ -1,21 +1,41 @@
 // @flow
 
+type ProxiedErrorValues = {|
+  +statusCode: number | string,
```
jaroslav-kubicek: I see `number | string` - what about _just_ string? `403` vs `"403"` => I would never be sure

mrtnzlml: Fixed. It was always like that (because it's proxied) but it's probably wrong. New `extensions` are always returning a string from now on.

mrtnzlml: Also:

> 403 vs "403" => I would never be sure

You should never use this error manually anyway...

jaroslav-kubicek:
rebase pls :)

mrtnzlml:
@jaroslav-kubicek Rebased.

PR #222, opened 2018-06-19
BC break: remove input arguments for whitelabel services

mrtnzlml:
Using correct dates is fairly complex and I don't want to leave it on the client at this moment. This is probably not going to be enough in the future but should be alright for now...


src/booking/resolvers/WhitelabelCarRental.js
```diff
@@ -0,0 +1,70 @@
...
+      dropoff: addHours(inboundArrival, ONE_WEEK),
+    });
+  } else if (booking.type === 'BookingMulticity') {
+    const trips = booking.trips || [];
+    trips.forEach(trip => {
+      const arrivalTime = idx(trip, _ => _.arrival.when.utc) || new Date();
+
+      return locationCodes.set(lastLocationCode(trip.legs), {
+        pickup: arrivalTime,
+        dropoff: addHours(arrivalTime, ONE_WEEK), // should be actually next departure of the next trip
```
RobinCsl: Due to your comment, this is still WIP?

src/booking/resolvers/WhitelabelCarRental.js
```diff
@@ -0,0 +1,70 @@
...
+  booking,
+}: AncestorType): Promise<CarRentalServiceType> => {
+  // we are interested in the IATAs at the end of the trips (arrivals only, no transfers)
+  // car rental is then interesting only for the time you are outside of the airport
+
+  const ONE_WEEK = 7 * 24;
+  const locationCodes = new Map();
+
+  if (booking.type === 'BookingOneWay') {
+    const arrivalTime = idx(booking, _ => _.arrival.when.utc) || new Date();
```
RobinCsl: Why do you default to `new Date()` if there is no date available? Would it not be better to throw an Error? I probably don't want to be given the option to rent a car with a wrong date -- it would annoy me and cause friction.

src/booking/types/outputs/services/CarRentalService.js
```diff
@@ -7,11 +7,13 @@ import CarRentalServiceRelevantCities, {
 } from './CarRentalServiceRelevantCities';
 import { type GraphqlContextType } from '../../../../common/services/GraphqlContext';
 
-export type CarRentalServiceType = {|
-  +locationCodes: $ReadOnlyArray<string>,
-  +pickup: Date,
-  +dropoff: Date,
-|};
+export type CarRentalServiceType = Map<
```
RobinCsl: This is a way of reorganising your data for the `locationCode` to be closer to their corresponding `pickup` and `dropoff`, correct?
(Also, the way it was done before was assuming there was only one `pickup` and one `dropoff`?)

RobinCsl: If I understand correctly, before you had: 
```
const carRentalService = {
  locationCodes : ['AA', 'BB', 'AB'],
  pickup: someDate,
  dropoff: someOtherDate,
}
```
when now it will look like:
```
const carRentalService = new Map();
Map.set('AA', {pickup: someDateForAA, dropoff: someOtherDateForAA});
Map.set('BB', {pickup: someDateForBB, dropoff: someOtherDateForBB});
etc.
```

src/booking/types/outputs/services/WhitelabeledServices.js
```diff
@@ -21,25 +20,22 @@ export default new GraphQLObjectType({
...
-      args: {
-        departureTime: {
-          type: GraphQLNonNull(GraphQLDateTime),
-        },
-      },
-      resolve: async ({ booking }: AncestorType, args) => {
+      resolve: async ({ booking }: AncestorType) => {
         // we are interested in every unique IATA of the airport
-        const iataSet = new Set();
+        const iataMap = new Map();
```
RobinCsl: Is the uniqueness property not lost?

src/booking/types/outputs/services/WhitelabeledServices.js
```diff
@@ -21,25 +20,22 @@ export default new GraphQLObjectType({
...
+
         booking.legs.map(({ departure, arrival }) => {
-          iataSet.add(departure.where.code);
-          iataSet.add(arrival.where.code);
+          const departureDate = idx(departure, _ => _.when.utc) || new Date();
```
RobinCsl: Same question about defaulting to `new Date()`.

src/booking/types/outputs/services/WhitelabeledServices.js
```diff
@@ -21,25 +20,22 @@ export default new GraphQLObjectType({
   fields: {
     lounge: {
       type: LoungeService,
-      args: {
-        departureTime: {
-          type: GraphQLNonNull(GraphQLDateTime),
-        },
-      },
-      resolve: async ({ booking }: AncestorType, args) => {
+      resolve: async ({ booking }: AncestorType) => {
```
RobinCsl: You are removing the `args.departureTime` because this information is already contained in the booking, correct?

src/booking/types/outputs/services/WhitelabeledServices.js
```diff
@@ -53,23 +49,15 @@ export default new GraphQLObjectType({
...
+      resolve: ({ booking }: AncestorType) => {
         // we are interested only in the IATA of the departure (you are leaving your car there)
         const iataCode = booking.departure.where.code;
 
         if (ParkingServiceAvailability[iataCode] === true) {
           return {
             iataCode: iataCode,
-            fromDate: args.fromDate,
-            toDate: args.toDate,
+            fromDate: idx(booking, _ => _.departure.when.utc) || new Date(),
```
RobinCsl: Same `new Date()` question.

src/booking/resolvers/WhitelabelCarRental.js
mrtnzlml: I don't know what's better here. It returns white label URL at the end so the idea was to construct at least something even if with the wrong dates so the user can correct them (instead of nothing). The dates are going to be mostly wrong anyway because we are setting only the default values here...

src/booking/resolvers/WhitelabelCarRental.js
mrtnzlml: shippable WIP 😊

src/booking/types/outputs/services/CarRentalService.js
mrtnzlml: Correct. I assumed there is only one set of the dates to work with. I am fixing it now... :)

src/booking/types/outputs/services/WhitelabeledServices.js
mrtnzlml: correct

This may change in the future because the requirements are more complex for other parts of the application but this is good enough for now.

src/booking/types/outputs/services/WhitelabeledServices.js
RobinCsl: Ok, actually, the Set object has same `keys()` and `values()` function. When Map actually has different `keys()` and `values()` functions, returning as expected resp. the keys, the values of the Map object.

src/booking/resolvers/WhitelabelCarRental.js
mrtnzlml: The user has to click the button and he'll get the opportunity to choose for which part of the trip he needs the car. Then we open the white label with default values. We try to set the default values as he would expect but there are cases where we really don't know. (You arrived to the destination - for how long do you need the car? We say 7 days by default but the real number is unknown.)

Long story short: I don't want to throw an error here. I want to show at least something even though it's maybe wrong.

RobinCsl: Got it, then it makes sense to me to do what you are doing here. :+1:

src/booking/resolvers/WhitelabelCarRental.js
mrtnzlml: I write a comment in the code. :trollface: 

Look, it's not that important to spend hours trying to make it perfect. I try to make it work first and if it's shitty I'll improve it later (or when I will be working with this code again). There is a chance that nobody will ever care about this...

src/booking/types/outputs/services/WhitelabeledServices.js
RobinCsl:
```
> Map.prototype.values == Map.prototype.keys
false
> Set.prototype.values == Set.prototype.keys
true
```

mrtnzlml: See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/values

> The keys() method is an alias for this method (for similarity with Map objects); it behaves exactly the same and returns values of Set elements.

src/booking/resolvers/WhitelabelCarRental.js
RobinCsl: I see, just asking to understand how you do things :)

mrtnzlml: Just a small note: people sometimes tend to create issues for this. I don't do that. The thing is that they are creating these issues but never actually fixing them. This way you have hundreds of void issues hiding the real issues. So I actually consider it to be a worse situation.

This comment is not such a big deal and I am just admitting that I am aware of this edge case but I am lazy to fix it now because I don't consider it important now.

RobinCsl: Got you. Thanks :)

PR #224, opened 2018-06-20
Remove key '_proxy' from the error response

mrtnzlml:
Closes: https://github.com/kiwicom/graphql/issues/212


RobinCsl:
But they can be useful sometimes? So now they will fail silently?

mrtnzlml:
@RobinCsl See: https://github.com/kiwicom/graphql/issues/212 (it's replaced with the new format)

RobinCsl:
Oh I see, so that can now be merged. Approved.

PR #225, opened 2018-06-20
Lounge service: do not throw 404 error

mrtnzlml:
404 actually means 200 in this case - just a stupid API


RobinCsl:
But that also means that if API is not available at all, we will fail silently. Correct?

PR #226, opened 2018-06-20
Adding location field to BookedFlightTimelineEvent

RobinCsl:


src/booking/types/outputs/BookingTimelineEvent.js
```diff
@@ -32,7 +33,14 @@ export default TimelineEvent;
 
 export const BookedFlightTimelineEvent = new GraphQLObjectType({
   name: 'BookedFlightTimelineEvent',
-  fields: commonFields,
+  fields: {
+    ...commonFields,
+    location: {
+      type: RouteStop,
+      description: 'Location of arrival',
```
mrtnzlml: Is the description correct?

RobinCsl: The `BookedFlight` should display: "_Booked flight to \_\_destination\_\__", so that's why we need the location of arrival.

PR #227, opened 2018-06-21
add original numeric id field to FAQArticle and FAQCategory

vborszeki:


src/FAQ/dataloaders/getFAQCategory.js
```diff
@@ -6,7 +6,7 @@ import { get } from '../../common/services/HttpRequest';
 import type { FAQCategoryType } from '../types/outputs/FAQCategory';
 
 export type FAQArticleItem = $ReadOnly<{|
-  id: string,
+  id: number,
```
git-toni: You should rather keep the `id` as a string and add a new field to this type called `databaseId` with number type.

src/FAQ/queries/__test__/FAQCategory.test.js
```diff
@@ -15,15 +15,16 @@ describe('allFAQCategories', () => {
     const resultsQuery = `query FAQSubcategories($id: ID!) { 
       FAQCategory(language: en, id: $id) {
         id
+        originalId
```
git-toni: The "consensus" in the repo is to call this field `databaseId` if I'm not mistaken. Pls rename this occurrence and all the others in the diff :slightly_smiling_face:

vborszeki: So which version should we stick with then?

jaroslav-kubicek: `originalId` is fine IMHO :)

jaroslav-kubicek:
Looks great, thanks. Pls just rebase, thx @vborszeki

src/FAQ/queries/__test__/FAQCategory.test.js
mrtnzlml: `id(unmasked: true)` 😊

git-toni: > Hotels have originalId, Booking has databaseId so in fact there is no consensus
Ouch. :smile:

src/FAQ/dataloaders/getFAQCategory.js
git-toni: :man_facepalming:

Issue #204, opened 2018-06-11
[Initial setup] Explain how to set up .env file


RobinCsl:
Explain the Readme where to get the information/who to contact regarding the development .env file.

PR #228, opened 2018-06-22
Readme: add installation instructions

mrtnzlml:
Closes: https://github.com/kiwicom/graphql/issues/204


README.md
```diff
@@ -66,10 +67,6 @@ _TODO: we should run tests there as well, we should also test the build!_
 2. build script will build the app into `.build` directory
 3. this directory and `node_modules` are packed together into the Docker image and executed in production
 
-# GraphiQL
```
RobinCsl: Why do you remove this part about GraphiQL? I believe it can be useful.

mrtnzlml: GraphiQL is there by default (after `yarn start`). It's not explained here at all but the only reason why is this necessary is to send custom auth headers. However, this is not a solution. I would like to do this:

1. move here our [customized GraphiQL](https://kiwi-graphiql.now.sh/) where you can log-in, or
2. start here [graphql-playground](https://github.com/prismagraphql/graphql-playground) which will allow you to send auth headers without additional tools

RobinCsl: I see. Maybe moving the customized GraphiQL would be the easiest for setup, right?

mrtnzlml: @RobinCsl Yes, I will start with this. There are many projects we should move to this repo and opensource. This is one of them. On it...

PR #229, opened 2018-06-22
Open-source customized GraphiQL interface

mrtnzlml:
This is just an import of our very old internal implementation...


RobinCsl:
Following the README was not enough: I had to execute `./scripts/build.sh` to actually see the GraphiQL interface.
When I try to login, it gives me the following error message in the console:
POST http://127.0.0.1:3000/undefined 404 (Not Found)
So it looks like the path dealing with login is not defined.

PR #235, opened 2018-06-25
Get rid of config file in FAQ queries

jaroslav-kubicek:
Get rid of deprecated `config/application.js` in all FAQ queries.


mrtnzlml:
@jaroslav-kubicek Can you please move it to the ENVs and set the ENVs on the server? This is a new way how we are going to do that (requested by DevOps).

jaroslav-kubicek:
any suggestion how to solve params in the middle with ENV? e.g. `https://api.skypicker.com/knowledgebase/api/v1/articles/${faqArticleId}/comment`

mrtnzlml:
@jaroslav-kubicek 

1. we can keep exactly this format: `https://api.skypicker.com/knowledgebase/api/v1/articles/${faqArticleId}/comment`, or maybe better
2. we should add there only the base URL so they can exchange it for a different one in a different region for example

(I have to ask again what is the main motivation actually)

PR #237, opened 2018-06-27
Whitelabel URL hotfix after response change

michalsanger:


michalsanger:
@mrtnzlml it was just NPM failure. How can I re-run the test on CI without the SSH option that just hangs and wait for timeout? :-(

mrtnzlml:
Dunno. There used to be a button for it (rebuild without cache) but I cannot find it now. Merging...

PR #238, opened 2018-06-27
Add boarding passes

kantord:


src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -110,6 +110,7 @@ export function sanitizeDetail(apiData: Object): Booking {
     assets: {
       ticketUrl: apiData.assets.eticket,
       invoiceUrl: apiData.assets.invoice,
+      boardingPasses: apiData.assets.boarding_passes,
```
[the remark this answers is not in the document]
kantord: hmm actually I was thinking the same, but I ended up copying the structure of the code already present there :-) Yeah I think it's a good idea to use idx in all of those nested values

src/booking/types/outputs/BookingAssets.js
```diff
@@ -19,5 +20,16 @@ export default new GraphQLObjectType({
       description: 'URL of the invoice.',
       resolve: ({ invoiceUrl }: BookingAssets): string => invoiceUrl,
     },
+
+    boardingPasses: {
+      type: new GraphQLList(BoardingPass),
+      description: 'Boarding passes for flights in this booking',
+      resolve: ({ boardingPasses }): BookingAssets =>
+        boardingPasses &&
```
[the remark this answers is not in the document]
kantord: Good idea! I didn't think of that. Thanks :-)

src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
```diff
@@ -42,6 +42,7 @@ Object {
     },
   },
   "assets": Object {
+    "boardingPasses": undefined,
```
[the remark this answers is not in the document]
kantord: Yeah sure :-)

kantord:
Everything should be resolved now (no pun intended)

src/booking/types/outputs/BookingAssets.js
```diff
@@ -11,13 +12,28 @@ export default new GraphQLObjectType({
...
     },
 
     invoiceUrl: {
       type: GraphQLString,
       description: 'URL of the invoice.',
-      resolve: ({ invoiceUrl }: BookingAssets): string => invoiceUrl,
+      resolve: ({ invoiceUrl }: BookingAssets): ?string => invoiceUrl,
+    },
+
+    boardingPasses: {
```
michalsanger: Last thing I'm missing from being totally happy is having this resolver tested in GraphQL query like here for example `src/booking/queries/__tests__/Booking.baggage.test.js`

kantord: Done! :+1:  https://github.com/kiwicom/graphql/pull/238/files#diff-c0af618332f61344b386641f60aa5de4R24

src/booking/datasets/booking-2707251.json
```diff
@@ -253,6 +253,10 @@
       "315289498": null,
       "all": null
     },
+    "boarding_passes": {
+      "315289498": null,
```
michalsanger: It's better to put some dummy string value then `null` so in the snapshot it's clear it works and it's not a failure.

kantord: :+1:

kantord:
Should be good now? :-)

PR #242, opened 2018-06-29
Add boarding passes to Leg

kantord:


src/flight/types/outputs/Leg.js
```diff
@@ -106,5 +108,23 @@ export default new GraphQLObjectType({
...
+        { id, bookingId }: { id: string, bookingId: string },
+        _,
+        { dataLoader }: GraphqlContextType,
+      ) => {
+        const booking = await dataLoader.booking.load(bookingId);
+        const boardingPasses = idx(booking, _ => _.assets.boardingPasses);
+
+        return {
+          flightNumber: id,
+          boardingPassUrl: boardingPasses && boardingPasses[id],
```
tbergq: how about moving this ti idx
`const boardingPassUrl = idx(booking, _ => _.assets.boardingPasses[id]);`

kantord: :man_shrugging: dunno. The your idx version looks fine to me, however I don't see the added value if there's no further nesting. We can add it if you want :man_shrugging:

tbergq: The added value would be that you don't have to do `boardingPasses && boardingPasses[id]` after using idx,  you can just do this really

```
return {
        flightNumber: id,
        boardingPassUrl: idx(booking, _ => _.assets.boardingPasses[id]);
}
```

tbergq:
Also flight query crashes if you add boarding pass to legs

kantord:
Crash fixed an tests added

PR #243, opened 2018-06-29
Nest leg under boarding pass

kantord:


src/booking/Booking.js
```diff
@@ -83,4 +83,5 @@ export type BookingAssets = {
   ticketUrl: ?string,
   invoiceUrl: ?string,
   boardingPasses: { [string]: string },
+  legs: Array<Leg>,
```
jaroslav-kubicek: => `Leg[]`

src/booking/types/outputs/BoardingPass.js
```diff
@@ -13,5 +14,10 @@ export default new GraphQLObjectType({
       type: GraphQLString,
       resolve: ({ boardingPassUrl }) => boardingPassUrl,
     },
+    leg: {
+      type: Leg,
+      resolve: ({ legs, flightNumber }) =>
+        legs.filter(leg => leg.id === flightNumber)[0],
```
jaroslav-kubicek: the case when an array can be empty after filtering should be handled.

kantord: `[][0]` returns `undefined`, which is exactly what we want so I think it is handled implicitly. I added a default value just to make is explicit

src/booking/types/outputs/BookingAssets.js
```diff
@@ -26,13 +26,15 @@ export default new GraphQLObjectType({
...
         boardingPasses,
+        legs,
       }: BookingAssets): Array<{
         boardingPassUrl: mixed,
         flightNumber: string,
       }> =>
         Object.entries(boardingPasses).map(([key, value]) => ({
           boardingPassUrl: value,
           flightNumber: key,
+          legs: legs,
```
tbergq: `legs: legs,` -> `legs,`

src/booking/types/outputs/BoardingPass.js
```diff
@@ -13,5 +14,10 @@ export default new GraphQLObjectType({
       type: GraphQLString,
       resolve: ({ boardingPassUrl }) => boardingPassUrl,
     },
+    leg: {
+      type: Leg,
+      resolve: ({ legs, flightNumber }) =>
+        legs.filter(leg => leg.id === flightNumber)[0] || null,
+    },
```
mrtnzlml: Everyone is now talking about how we should use the Glossary and we are thinking about how to migrate our GraphQL. It didn't go very well here... 🙄 😬 (https://kiwi.wiki/#/comms/glossary)

kantord: So we should rename this to `segment`?

mrtnzlml: We will have to do it somehow systematically later I guess...

PR #245, opened 2018-07-02
Eslint add camelcase rule

tbergq:


src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -72,7 +72,7 @@ export function sanitizeListItem(apiData: Object): BookingsItem {
 
 export function sanitizeDetail(apiData: Object): Booking {
   const common = sanitizeListItem(apiData);
-  const { bag_params } = apiData;
+  const { bag_params: bagParams } = apiData;
```
michalsanger: Hm, this is strange to me. Rename response prop just to make some ESLint rule happy? Isn't better to ignore the rule per line or per file in cases like sanitizer?

tbergq: In that case I would actullay prefer to do `height: apiData.bag_params.hand_height,`

tbergq:
@michalsanger 🚀

src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -83,27 +82,33 @@ export function sanitizeDetail(apiData: Object): Booking {
       ),
       cabin: [
         {
-          height: bag_params.hand_height,
-          length: bag_params.hand_length,
-          width: bag_params.hand_width,
-          weight: bag_params.hand_weight,
-          note: bag_params.hand_note === '' ? null : bag_params.hand_note,
+          height: apiData.bag_params.hand_height,
```
michalsanger: Best if `idx` was used

tbergq: why? this code is exactly as it was before. If it was safe not to use idx before, this change has not changed that

michalsanger: That's true but what about my OCD once I've seen this? :-)

tbergq: I would say that is out of scope if this PR. This file is 250 lines without idx, and I guess we have countless other files not using idx. 

I am just trying to prevent guys from using snake_case here. 🙏

michalsanger: OK, do the rebase plz

tbergq: @michalsanger 🚀

PR #246, opened 2018-07-02
Eslint add cury all rule

tbergq:


.eslintrc
```diff
@@ -21,7 +21,7 @@
     "flowtype",
   ],
   "rules": {
-    "curly": "error",
```
mrtnzlml: What was the default behavior actually?

michalsanger: @mrtnzlml plz merge if the answer is not a blocker for you

tbergq: "all" is supposed to be the default, but maybe some plugin or whatever overwrites this. 🤷‍♂️

PR #249, opened 2018-07-04
Booking: Add boarding pass available date

tbergq:


src/booking/Booking.js
```diff
@@ -79,9 +79,16 @@ export type Booking = BookingsItem & {
   contactDetails: ContactDetails,
 };
 
+export type BoardingPass = {|
+  boardingPassUrl: mixed,
+  flightNumber: string,
+  availableAt: ?string,
+  leg: ?Leg,
+|};
```
mrtnzlml: Let's go read-only here... :)

tbergq: 🚀

src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -16,12 +21,14 @@ import type { InboundOutboundData } from '../types/outputs/BookingReturn';
...
 export function sanitizeListItem(apiData: Object): BookingsItem {
   const bid = parseInt(apiData.bid);
+
   const legs = apiData.flights.map((flight): Leg => ({
     id: flight.id,
     bookingId: bid,
     recheckRequired: flight.bags_recheck_required,
     isReturn: flight.return === 1,
     flightNo: flight.flight_no,
+    boardingPassAvailableAt: idx(flight, _ => _.ticket_available_at),
```
mrtnzlml: Not now but: maybe it's about time to stop using `idx` in this repo? Optional chaining is [in stage 2](https://github.com/tc39/proposal-optional-chaining) and there is already Babel plugin for it.

tbergq: Yes, I read about it, we could definitely implement this, in another PR 😄

src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
```diff
@@ -42,11 +42,26 @@ Object {
...
   },
   "assets": Object {
-    "boardingPasses": Object {
-      "SP3237959_4825962": null,
-      "SP3237959_4825963": "https://very.printable/pdf",
-      "all": null,
-    },
+    "boardingPasses": Array [
+      Object {
+        "availableAt": undefined,
```
mrtnzlml: Just to be sure - is this `undefined` expected?

src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
```diff
@@ -300,6 +319,7 @@ Object {
           "code": "WAW",
         },
       },
+      "boardingPassAvailableAt": undefined,
```
mrtnzlml: Same question here - sometimes it's `undefined`, sometimes `null`. Seems dodgy...

src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
tbergq: well, the json file that is used to generate this file does not contain any field called `ticket_available_at` and since we access it with idx it will then be undefined. Why this particular file does not have this field, I do not know.

So I would say it is expected here

src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
tbergq: undefined because same as over, it is null when the field exists on the response, it could be null, if it is never available or a date string if it will be available

src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
tbergq: if course we could do idx(...) || null if you feel it is more consistent

mrtnzlml: So maybe the right answer would be to update our JSON datasets?

src/booking/dataloaders/__tests__/__datasets__/booking.detail.json
```diff
@@ -259,9 +259,9 @@
       "SP1123103_1678073": null
     },
     "boarding_passes": {
-      "SP3237959_4825962": null,
-      "SP3237959_4825963": "https://very.printable/pdf",
-      "all": null
+      "SP1123103_1678074": null,
+      "all": null,
+      "SP1123103_1678073": null
     },
```
tbergq: Inspecting the result with reset let client, I saw that it now returns "boardingPass" and "boarding_pass" with the same data, we use the latter. 

boarding_passes is this file was added manually and did not have the correct data

mrtnzlml: Do I understand correctly we had invalid JSON dataset? Is everything fine now?

tbergq: Yes, there was manually added to this file 
https://github.com/kiwicom/graphql/pull/238/files#diff-a691f0afe73b504bec6d0854f6397147R261

and boarding_pass did not have same data as boardingPass, as one would expect.

could be other flaws in this file as well 🤷‍♂️

PR #251, opened 2018-07-04
AirHelp+ offer MVP

michalsanger:


dinodsaurus:
Looks good👍, Im only not sure if we have `bookingId` but I will check that 🤔

PR #252, opened 2018-07-04
Adding NoMoreEdits event

RobinCsl:


src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -57,6 +57,8 @@ export function sanitizeListItem(apiData: Object): BookingsItem {
   } else if (type === 'BookingMulticity') {
     additionalFields.trips = createTrips(apiData.segments, legs);
   }
+  const allowedToChangeFlights =
+    idx(apiData, _ => _.config.allowedToChange.flights) || 48;
```
tbergq: Why is 48 default value, what does it mean?

RobinCsl: It was not very clear either in the API, but it should be the number of hours prior departure where it is allowed to change <insert config element here>.
So in case it is missing, I thought that 2 days before the flight was the limit for flights to be changed. But you're right, I should have confirmation from someone who knows for sure.

RobinCsl:
@tbergq Updated; after discussions on Slack with PM, behaviour should be:
- `config.allowed_to_change.flights == null`:  no `NoMoreEdits` event;
- else, `NoMoreEdits` generated with given value of `config.allowed_to_change.flights`.

src/booking/resolvers/BookingTimeline.js
```diff
@@ -302,3 +307,20 @@ export function generateTransportFromAirportEvent(
...
+
+export function generateNoMoreEditsEvent(booking: Booking): ?NoMoreEditsType {
+  const departureTime = idx(booking.departure, _ => _.when.local);
+  const allowdToChangeFlights = idx(booking, _ => _.allowedToChangeFlights);
+  if (departureTime && allowdToChangeFlights) {
+    const noMoreEditsTime = DateTime.fromJSDate(departureTime, {
+      zone: 'UTC',
+    })
+      .minus({ hours: allowdToChangeFlights })
+      .toJSDate();
```
tbergq: don't the backend already return `when.utc` do you need to take the local and convert it to utc?

RobinCsl: As far as I understand, the local times are also given in UTCZ, i.e. the information about the timezone is lost; and in many places where `DateTime.fromJSDate` is used, the option `{zone: 'UTC' }` is used so I believed it was better practice. Please correct me if I am doing this wrong.

PR #254, opened 2018-07-09
Making terminal accessible from API data; modified BoardingTimelineEv…

RobinCsl:
…ent to make terminal accessible

Note: `terminal` is available in the API response **only when querying ONE booking**.

e.g.
Querying through all bookings
```
...
"where": {
"city_id": "hong-kong_cn",
"code": "HKG",
"country": "CN",
"name": "Hong Kong",
"station": "Hong Kong International"
}
...
``` 

Querying with specific booking id: 
```
"where": {
"city_id": "hong-kong_cn",
"code": "HKG",
"country": "CN",
"gate": null,
"name": "Hong Kong",
"station": "Hong Kong International",
"terminal": "1"
}
```


src/booking/queries/__tests__/__snapshots__/BookingTimeline.test.js.snap
```diff
@@ -76,7 +76,7 @@ Object {
         },
         Object {
           "__typename": "BoardingTimelineEvent",
-          "gate": "gate number",
+          "terminal": "",
```
tbergq: Could you also test here with a dataset that contains a terminal so we know it works?

michalsanger: Shouldn't `null` be returned when the value is not known?

RobinCsl:
> I consider this as bad design of REST API that we don't have to copy into GraphQL

I will try to work around it. That means that the Bookings resolver should have an intermediate step where we first gather the `bid`'s from the current implementation and then load each booking through the booking dataloader. Does that seem reasonable?

FYI: 
- number of child attributes for booking object in Bookings loader: 38
- number of child attributes for booking object in Booking loader: 62

tbergq:
> That means that the Bookings resolver should have an intermediate step where we first gather the bid's from the current implementation and then load each booking through the booking dataloader

We already do this for allowedBaggage and assets(probably more fields)

And this will only be triggered if we actually ask for that particular field in the query

michalsanger:
@RobinCsl that's one of the benefits when using Dataloader. It handles the intermediate step of gathering `bid`'s.

RobinCsl:
@michalsanger Yes, I understand. Thank you :) 
@tbergq So if I was to do a similar thing as for assets, since terminal is _"buried"_ inside `RouteStop`, which is not lying directly under `BookingInterface`, I will need to either pass down the `bid` to RouteStop (which is not elegant), or I will need to use the `bid` to retrieve the parents of `RouteStop`, which can be found in `Trip` for instance. That will imply a lot of modifications I believe. Thoughts?

tbergq:
Maybe we are doing to much work in our ApiSanitizer here, and not using our types correctly. What do you think @michalsanger?

I do see that it is difficult to get terminal from `customerBookings` with our current code. 

Though since terminal field is accessible from `customerBookings` we need to have a correct return value.

src/booking/datasets/booking-item-multicity-4903131.json
```diff
@@ -1,4 +1,7 @@
 {
+  "additional_bookings": {
+    "details": []
+  },
```
RobinCsl: I had to add these to two datasets because of the way the tests were structured. Somehow, it seems to me these files should be more similar to `booking-2707251.json` than to `AllBookings.json`... Thoughts?

src/booking/datasets/booking-item-return-3222550.json
```diff
@@ -1,4 +1,7 @@
 {
+  "additional_bookings": {
+    "details": []
+  },
```
RobinCsl: Same comment here.

src/booking/types/outputs/BookingReturn.js
```diff
@@ -26,12 +27,26 @@ const BookingReturn = new GraphQLObjectType({
...
     inbound: {
       type: Trip,
       description: 'Return trip back from destination to origin.',
-      resolve: ({ inbound }: BookingReturnData): ?TripData => inbound,
+      resolve: async (
+        { id }: BookingReturnData,
+        args: Object,
+        { dataLoader }: GraphqlContextType,
+      ): Promise<?TripData> => {
+        const { inbound } = await dataLoader.booking.load(id);
```
mrtnzlml: This way we can probably remove the second bookings dataloader. What do you think? However, I am really afraid to merge this, because BE is not able to handle it...

RobinCsl: Well, you still need a way to get all the `bid`'s of the customer. So we might not want to remove it altogether. 
And yes, it increases significantly the number of requests whenever you ask for any data related  to the itinerary. But at the same time, if you ask for `assets`, `allowedBaggage` or `passengers`, it will not be more work. So it depends how many queries there are which contain fields related to the itinerary and which don't call `assets, `allowedBaggage` or `passengers`.
Thoughts?

mrtnzlml: I think it would be better to do it the same ways with assets or baggage.

RobinCsl:
Okay, I believe now it is the closest it can be to `terminal` and it does not trigger unnecessary calls.
The only downside is the passing around of the `bid`. 

Note: To identify which flight it is that I want to know the terminal of, I use the timestamp of the flight which should be unique across the booking.

PR #255, opened 2018-07-10
Add the quantity of checked baggage

kantord:


tbergq:
It works for 99% of my test bookings, however bid 6776372 has one checked bag, but graphql says 0

kantord:
Why :cry: :sob:

kantord:
2018-07-10-130418_726x603_scrot

Probably this is the explanation

kantord:
Closing because we have a different plan now, and we'll not need this exact modification. If you need it, feel free to re-open.

PR #257, opened 2018-07-11
Add vehicle, operatingAirline and pnr fields to Leg

vborszeki:


src/booking/dataloaders/__tests__/__snapshots__/ApiSanitizer.test.js.snap
```diff
@@ -192,7 +228,19 @@ Object {
...
         "isReturn": false,
+        "operatingAirline": Object {
+          "iata": "LO",
+          "name": "LOT Polish Airlines",
+        },
+        "pnr": "test",
         "recheckRequired": false,
+        "vehicle": Object {
+          "manufacturer": null,
+          "meal": null,
```
kantord: It'd be nice to have an example of an actual airplane model in the tests

src/flight/types/outputs/OperatingAirline.js
```diff
@@ -0,0 +1,19 @@
...
+
+import { GraphQLObjectType, GraphQLString } from 'graphql';
+
+import type { OperatingAirline } from '../../Flight';
+
+export default new GraphQLObjectType({
+  name: 'OperatingAirline',
+  fields: {
+    iata: {
+      type: GraphQLString,
```
kantord: It'd be nice to have a `description` field for some of the GraphQL field, for example `iata` is not immediately clear for me

src/flight/Flight.js
```diff
@@ -54,3 +57,13 @@ export type Airline = {|
   code: string,
   isLowCost: boolean,
 |};
+
+export type OperatingAirline = {|
+  iata: string,
```
tbergq: could you make these types read-only

src/flight/Flight.js
```diff
@@ -54,3 +57,13 @@ export type Airline = {|
...
+  name: string,
+|};
+
+export type Vehicle = {|
+  manufacturer: string,
+  model: string,
```
tbergq: also these

PR #258, opened 2018-07-11
Adding BagDrop and BagPickup events

RobinCsl:


src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -124,6 +124,7 @@ export function sanitizeDetail(apiData: Object): Booking {
         },
       ],
     },
+    numberCheckedBaggage: apiData.bags,
```
tbergq: this is apparently not that easy, there is also another PR for this #255

RobinCsl:
Closing the PR while waiting for new API endpoint to be used by @kantord.

PR #267, opened 2018-07-13
Fix multicity type of bookings when segments are incorrect

### LINKED ISSUE #266: Fix buggy multicity
For some bookings, mainly in testing accounts, query fails with unexpected error for multicity.

jaroslav-kubicek:
Fixes #266 

Be more defensive and expect that segment indices returned from REST API might not be 100 % correct and especially the last index in a list might refer to the empty segment.


mrtnzlml:
@jaroslav-kubicek Did you report the bug somewhere so they can fix it?

PR #269, opened 2018-07-16
Adding TimeToCheckin and CheckinClosing events

RobinCsl:


src/booking/resolvers/BookingTimeline.js
```diff
@@ -350,3 +360,44 @@ export function generateEnterDetailsEvent(booking: Booking): ?EnterDetailsType {
...
+export function generateTimeToCheckinEvent(
+  booking: Booking,
+): ?TimeToCheckinType {
+  const departureTime = idx(booking.departure, _ => _.when.local);
+  const onlineCheckinIsAvailable = idx(
+    booking,
+    _ => _.onlineCheckinIsAvailable,
+  );
+  if (onlineCheckinIsAvailable && departureTime) {
+    const timestamp = DateTime.fromJSDate(departureTime)
```
mrtnzlml: Does it work correctly with the right time or it takes the local machine time?

RobinCsl: I think you are referring to the missing option? It seems to be working fine without, but I will add it for consistency with the other instances of DateTime being used.

mrtnzlml: I would prefer to be more sure about that... 😬

src/booking/resolvers/__tests__/BookingTimeline.test.js
```diff
@@ -707,3 +709,69 @@ describe('generateEnterDetailsEvent', () => {
...
+        onlineCheckinIsAvailable: true,
+      }),
+    ).toEqual({
+      timestamp: new Date('2018-04-30T14:10:57.000Z'),
+      type: 'CheckinClosingTimelineEvent',
+    });
+  });
+  it('should return null if online checkin is not available or departure time is not available', () => {
+    expect(
+      // $FlowExpectedError: full Booking object is not needed for this test
```
mrtnzlml: Does it make sense to stop using Flow in the test files? Or at least make the `// @flow` optional?

src/booking/resolvers/BookingTimeline.js
RobinCsl: Well, on the examples I tested, I get the same time as what is displayed in the current app.

src/booking/resolvers/__tests__/BookingTimeline.test.js
RobinCsl: Would you prefer me to refactor the tests with some mock of a Booking type instead of using just a subset of properties for each test?

PR #272, opened 2018-07-17
Broke down big resolver file into smaller files; did the same to acco…

RobinCsl:
…mpanying test file


RobinCsl:
@mrtnzlml  That's right; I also reordered two events, but that's insignificant.

PR #275, opened 2018-07-19
Add singleBooking query

jalbertsr:


src/booking/dataloaders/SingleBooking.js
```diff
@@ -0,0 +1,35 @@
...
+import DataLoader from 'dataloader';
+import { get } from '../../common/services/HttpRequest';
+import Config from '../../../config/application';
+import { sanitizeDetail } from './ApiSanitizer';
+import type { Booking } from '../Booking';
+
+export type Args = {
+  simpleToken: string,
+  id: number,
+};
```
tbergq: could this be exact type, and read only?

src/booking/dataloaders/SingleBooking.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default function createSingleBookingLoader() {
+  return new DataLoader(
+    (bookings: $ReadOnlyArray<Args>) => batchLoad(bookings),
+    {
+      cacheKeyFn: key => parseInt(key, 10),
```
tbergq: This key is an object in this case right?
you should use `json-stable-stringify`

mrtnzlml:
So no `viewer`? That's a pity. 😞

tbergq:
We did not see how that would solve the problem 🤷‍♂️

mrtnzlml:
It's just a different query design. You would go from this:

```graphql
{
  booking(bid: ID)
  singleBooking(bid: ID, simpleToken: ID)
  somethingElse(simpleToken: ID)
}
```

To this:

```graphql
{
  viewer(simpleToken: ID) {
    booking(bid: ID)
    singleBooking(bid: ID)
    somethingElse
  }
}
```

And therefore it would be clear that `booking` vs. `singleBooking` is quite unfortunate and you can simplify (maybe with `bookingViewer`):

```graphql
{
  bookingViewer(simpleToken: ID) {
    booking(bid: ID) # singleBooking not needed now
    customerBookings # could be here as well - why not, viewer can have different auth methods as well :)
  }
}
```

It was just an idea. No big deal.

jaroslav-kubicek:
What about `simpleToken` argument to be sent in `Authorization` header. Looks like a little discrepancy to me `booking(id: ID)` with `Authorization` header vs `singleBooking(id: ID, simpleToken: string)`

tbergq:
We still send it as a query parameter to the api-endpoint 🤷‍♂️ 

Then also, how do we separate simple_token from a full token?

src/booking/queries/SingleBooking.js
```diff
@@ -0,0 +1,31 @@
...
+import GraphQLBooking from '../types/outputs/Booking';
+import type { GraphqlContextType } from '../../common/services/GraphqlContext';
+import type { BookingsItem } from '../Booking';
+
+export default {
+  type: GraphQLBooking,
+  description: 'Find booking by its id and its simple access token.',
+  args: {
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
```
jaroslav-kubicek: if it accepts only bid, then it should be `Int` - same as databaseId?

mrtnzlml:
@jaroslav-kubicek From what I understood "simple token" is not auth method as such but more like a URL parameter. I am afraid to use `Authorization` header for this because then there is going to be another auth method and we will have to add it again.

jaroslav-kubicek:
ok, I see, forget my note above regarding the simple token.

src/booking/queries/SingleBooking.js
tbergq: You are right, I missed that 😞 
cc @jalbertsr

PR #277, opened 2018-07-20
Ancillaries: Airhelp+ Removed bookingId

dinodsaurus:
I removed `bookingId` because it is not needed


mrtnzlml:
@dinodsaurus Hi! Seems like you wanted to remove the booking ID but you are stuck now because of the backward compatibility check. Correct? Do you know how to deal with it so you can actually remove it? I think it's safe to remove it even though it's BC break because it's still very fresh... :)

dinodsaurus:
Hey @mrtnzlml yea I wanted to remove i t😅  but I don't know how to make `circleci` pass if you could help me out that would be great 😄

mrtnzlml:
You can comment [this line](https://github.com/kiwicom/graphql/blob/master/scripts/CheckBackwardCompatibility.js#L57) and run `yarn tet-bc`. Please be careful with this and do it only when you are absolutely sure it's not harmful (like in this case)... :)

PR #278, opened 2018-07-20
Add update passenger mutation

tbergq:


config/application.js
```diff
@@ -27,6 +27,8 @@ export default {
         queryParameters,
       ),
     rates: 'https://api.skypicker.com/rates',
+    updatePassenger: (bookingId: string) =>
```
mrtnzlml: We would like to stop using this file because it's centralized and we have a feature based project. Could you please move it somewhere else?

tbergq: sure, just keep it in the resolver then?

mrtnzlml: Yes, that makes sense to me.

src/booking/mutations/UpdatePassenger.js
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    await patch(
+      config.restApiEndpoint.updatePassenger(bookingId),
+      payload,
+      headers,
+    );
+
+    return {
+      success: true,
```
mrtnzlml: There is no response from the server?

tbergq: nope, 204 no content 😞

src/booking/mutations/UpdatePassenger.js
```diff
@@ -0,0 +1,80 @@
...
+  args: {
+    id: {
+      type: GraphQLNonNull(GraphQLID),
+      description: 'The global booking id',
+    },
+    simpleToken: {
+      type: GraphQLString,
+    },
+    passengers: {
+      type: GraphQLNonNull(new GraphQLList(GraphQLPassengerInputType)),
```
mrtnzlml:
```diff
-      type: GraphQLNonNull(new GraphQLList(GraphQLPassengerInputType)),
+      type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLPassengerInputType))),
```

Makes sense? (and yes, you can skip the `new` 😊)

src/booking/mutations/UpdatePassenger.js
```diff
@@ -0,0 +1,80 @@
...
+|};
+
+export default {
+  type: GraphQLUpdatePassenger,
+  args: {
+    id: {
+      type: GraphQLNonNull(GraphQLID),
+      description: 'The global booking id',
+    },
+    simpleToken: {
```
mrtnzlml: This is what I was scared of - it's already going crazy like the "language" or "locale" inputs... 😄

I am still fine to accept it though.

tbergq: Maybe you are right with the `viewer`, but this is what we have at the moment 🤷‍♂️

src/booking/mutations/UpdatePassenger.js
```diff
@@ -0,0 +1,79 @@
...
+
+type PassengerInputType = {|
+  +passengerId: number,
+  +documentExpiry: Date,
+  +documentNumber: string,
+|};
+
+type Args = {
+  +id: string,
+  passengers: $ReadOnlyArray<PassengerInputType>,
```
mrtnzlml: Why is not `passengers` field read-only?

tbergq: I though it was not necessary with `$ReadOnlyArray` but that might be different?

tbergq:
@mrtnzlml 🚀

src/booking/mutations/UpdatePassenger.js
mrtnzlml: Sorry - I forgot to send an explanation. You have to make `+passenger` to prevent overwriting the array of PAX, then read-only array to prevent things like pushing new PAX into the array and then even the PAP itself to prevent modifications of the passenger. So correctly it's like this:

```
type Args = {|
  +id: string,
  +passengers: $ReadOnlyArray<{|
    +something: string,
  |}>
|}
```

I know, it can be annoying but it makes sense. See: https://goo.gl/M25Xvw

tbergq: Right 😄 it should be fixed now

PR #280, opened 2018-07-20
[Documents] possible to query online document urls with locale

RobinCsl:


src/RootQuery.js
```diff
@@ -57,5 +58,6 @@ export default new GraphQLObjectType({
     node: nodeField,
     airHelpPlusOffer: AirHelpPlusOffer,
     singleBooking: SingleBooking,
+    allDocuments: AllDocuments,
```
tbergq: Can we try to keep this list alphabetical please. It makes it so much easier to find queries in graphiql

mrtnzlml: @tbergq Time to implement automatic sort? 😬

RobinCsl: @tbergq Sorted. Not automatic though.

tbergq:
Ok, I see what I am not to familiar with this pattern. Could you have a look at it @mrtnzlml?

src/common/services/TestingTools.js
```diff
@@ -68,8 +68,15 @@ export class BookingComApiMock {
 export const graphql = async (
   query: string,
   variables: ?Object,
+  acceptLanguage: ?string,
```
mrtnzlml: Please try not to do that. This function should be compatible with `graphql` function from GraphQL itself. Or at least that was the original idea.

RobinCsl: Ok, I will try to find a different way to pass the language when creating the context value. 
What do you mean by 
> This function should be compatible with graphql function from GraphQL itself

? Because they do not have the same signature anyway.

mrtnzlml: Oh, that's right. So forget my comment, please.

PR #282, opened 2018-07-23
Nest pkpasses under boarding pass, deprecate pkpass under passenger

tbergq:
Connecting pkpass to passenger was a bad decision. I need access it under boarding pass, and rather attach passenger to the pkpass.


src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -166,18 +167,32 @@ export function sanitizeDetail(apiData: Object): Booking {
 function sanitizeBoardingPasses(
   boardingPasses: ?{ [string]: Object },
   legs: Leg[],
+  passengers: Array<Object>,
```
RobinCsl: Maybe using type `Passenger` would be better?

tbergq: Yes, I will look into it 👍

src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -166,18 +167,32 @@ export function sanitizeDetail(apiData: Object): Booking {
...
     const leg = legs.find(leg => leg.id === key);
+
     return {
       boardingPassUrl: value,
       flightNumber: key,
       availableAt: idx(leg, _ => _.boardingPassAvailableAt),
       leg,
+      pkpasses: passengers
+        .filter(passenger => {
+          return (
```
RobinCsl: I understand that you are deprecating the access of pkpasses under passengers; however, that means resolvers still need to make pkpasses available under passengers, and this way of getting pkpasses for boarding passes now seems strange because you did not need passengers before. Would it be possible to create another way to resolve the pkpasses without having to use passenger as a proxy to get it?

tbergq: No, because this is how the data is returned from the rest api:

```
passengers: [{..., pkpass}, {..., pkpass}]

```

tbergq:
@RobinCsl 🚀

PR #283, opened 2018-07-23
Adding insurance prices to booking

RobinCsl:
@tbergq Do you think I should add an entry to the response corresponding to the 'none' or 'no insurance' case (in which case it is free), instead of handling the special case in the app itself?


src/booking/Booking.js
```diff
@@ -102,3 +103,8 @@ export type BookingAssets = {
   boardingPasses: BoardingPass[],
   legs: Leg[],
 };
+
+export type InsurancePrice = {|
+  +type: 'travel_basic' | 'travel_plus',
```
tbergq: There are also deprecated types `skysilver` and `skygold`
https://github.com/kiwicom/graphql/blob/master/src/booking/types/outputs/Passenger.js#L65

RobinCsl: The way I resolve the insurance prices, I filter on the type and consider only the non-deprecated `travel_basic` and `travel_plus`. Is that a bad thing?

src/booking/types/outputs/Passenger.js
```diff
@@ -14,7 +14,7 @@ import type { Passenger } from '../../Booking';
 import GraphQLVisa from './Visa';
 import GraphQLPkpass from './Pkpass';
 
-const InsuranceTypeEnum = new GraphQLEnumType({
+export const InsuranceTypeEnum = new GraphQLEnumType({
```
tbergq: Maybe we should move this to a separate file, now that more types are using it?

src/booking/Booking.js
tbergq: 🤔 It is probably ok. I think you are right.

tbergq:
> Do you think I should add an entry to the response corresponding to the 'none' or 'no insurance' case (in which case it is free), instead of handling the special case in the app itself?

It sounds like a good idea yes.

src/booking/types/outputs/Passenger.js
RobinCsl: Ok, will do that.

src/booking/dataloaders/ApiSanitizer.js
```diff
@@ -347,3 +352,17 @@ function detectType(apiData): BookingType {
 
   return 'BookingOneWay';
 }
+
+function parseInsurancePrices(insurancePrice, currency): InsurancePrice[] {
+  const types = Object.keys(insurancePrice);
+
+  return types
+    .filter(type => type === 'travel_plus' || type === 'travel_basic')
```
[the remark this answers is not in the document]
RobinCsl: > 🤔 It is probably ok. I think you are right.

I will ignore this comment then 😉

RobinCsl:
@tbergq 🚀

PR #285, opened 2018-07-24
Correct Booking statuses

Strajk:
* NEW
* PENDING
* CANCELLED
* DELETED
* EXPIRED
statuses exist only on extras, not bookings themselves

~~Additionally, there's also "refunding" Booking status~~
`refunding` status exists only in db, but API transform it to `refunded` right before responding


jaroslav-kubicek:
Hola, this is breaking change, so normally we can't do that here.

But I guess these values you removed are never ever returned from booking api for any booking, is that correct?

If so => BC is aceptable => remove these statuses from schema snaphost: https://github.com/kiwicom/graphql/blob/master/scripts/graphql-schema-snapshot.json#L6260 so the script will think they never existed

mrtnzlml:
@jaroslav-kubicek We already sorted this out in person. @Strajk is going to deprecate these values.

jaroslav-kubicek:
I also noticed `BOOKING_STATUSES` on frontend includes `refunding` 🤔

Strajk:
@mrtnzlml i can and i will :)

@jaroslav-kubicek I will follow up with MRs to FE repo and Skymock that will remove 'refunding' status

PR #286, opened 2018-07-24
Deprecate section parameter for "FAQCategory" query

### LINKED ISSUE #276: Remove "section" param from FAQCategory query
- Remove "section" param from FAQCategory query - it was a silly idea 🙈 
- Do not return top-level category in `ancestors` field (as it's now) so the list doesn't contain "section" categories "Before booking", "After boooking"

First deprecate, release new version of SmartFAQ and then remove

jaroslav-kubicek:
Closes #276 

It was stupid.

Istead, I made root "section" categories (e.g. "After booking") invisible in subcategories.

### DIFF
--- a/src/FAQ/queries/FAQCategory.js
+++ b/src/FAQ/queries/FAQCategory.js
@@ -1,42 +1,10 @@
 // @flow
 
 import { GraphQLID, GraphQLNonNull } from 'graphql';
-import { fromGlobalId } from 'graphql-relay';
 
 import FAQCategory from '../types/outputs/FAQCategory';
 import FAQSection from '../types/enums/FAQSection';
 import type { GraphqlContextType } from '../../common/services/GraphqlContext';
-import type { FAQCategoryItem } from '../dataloaders/FAQCategories';
-import { sectionToCategories } from '../dataloaders/FAQCategories';
-
-const findCategory = (
-  categories: $ReadOnlyArray<FAQCategoryItem>,
-  categoryId: number,
-): FAQCategoryItem | null => {
-  const parentCategory = categories.find(c => c.id === categoryId);
-
-  if (parentCategory) {
-    return parentCategory;
-  }
-
-  for (const category of categories) {
-    const subcategories = category.subcategories || [];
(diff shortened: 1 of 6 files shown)


src/FAQ/queries/FAQCategory.js
```diff
@@ -30,6 +31,13 @@ const findCategory = (
   return null;
 };
 
+const sectionCategories = Object.values(sectionToCategories);
```
michalsanger: Could you use `new Set(Object.values(sectionToCategories))` so the filter on line 37 can be much more faster?

src/FAQ/queries/FAQCategory.js
```diff
@@ -42,26 +50,24 @@ export default {
...
-    { id, section }: Object,
+    { id }: Object,
     { dataLoader }: GraphqlContextType,
   ) => {
     const categoryId = Number(fromGlobalId(id).id);
-    const categories = await dataLoader.FAQCategories.load({
-      // dataloader needs to be called with value => null can't be used as default
-      section: section || 'all',
-    });
+    const categories = await dataLoader.FAQCategories.load({});
```
michalsanger: Looks hacky, could only `dataLoader.FAQCategories.load()` be used?

jaroslav-kubicek: nope, dataloader needs to called with value

michalsanger: Yeah, but you can wrap it https://github.com/kiwicom/graphql/blob/master/src/location/dataloaders/Location.js

src/FAQ/queries/FAQCategory.js
jaroslav-kubicek: fixed

src/FAQ/queries/FAQCategory.js
jaroslav-kubicek: okey, you win _co s tebou_

jaroslav-kubicek: refactored @michalsanger

PR #289, opened 2018-07-25
use singleBooking instead of booking

jalbertsr:
This is BC (no one is using this query), changed param `simpleToken` to `authToken` and used correct type `Int` instead of `ID`

Changing `booking` to `singleBooking` was needed because `singleBookingQuery` was failing loading fields that were using `booking` for example `assets`, `passengers`, `contactDetails`...

References https://github.com/kiwicom/smart-faq/issues/489

### DIFF
--- a/src/booking/dataloaders/SingleBooking.js
+++ b/src/booking/dataloaders/SingleBooking.js
@@ -3,12 +3,12 @@
 import DataLoader from 'dataloader';
 import stringify from 'json-stable-stringify';
 import { get } from '../../common/services/HttpRequest';
-import Config from '../../../config/application';
 import { sanitizeDetail } from './ApiSanitizer';
-import type { Booking } from '../Booking';
+import { queryWithParameters } from '../../../config/application';
+import type { BookingInterfaceData } from '../types/outputs/BookingInterface';
 
 export type Args = {|
-  +simpleToken: string,
+  +authToken: string,
   +id: number,
 |};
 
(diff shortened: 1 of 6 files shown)


src/booking/dataloaders/SingleBooking.js
```diff
@@ -24,13 +24,16 @@ export default function createSingleBookingLoader() {
...
   return responses.map(result => sanitizeDetail(result));
 }
 
-async function fetchFAQ(bid: number, simpleToken: string): Promise<Booking> {
-  return await get(Config.restApiEndpoint.singleBooking(bid, simpleToken));
+async function fetchFAQ(
+  bid: number,
+  authToken: string,
+): Promise<BookingInterfaceData> {
+  return await get(Config.restApiEndpoint.singleBooking(bid, authToken));
```
tbergq: from https://github.com/kiwicom/graphql/pull/278#discussion_r204027949
> We would like to stop using this file because it's centralized and we have a feature based project. Could you please move it somewhere else?

jaroslav-kubicek:
Since you started with BC, then I propose to return `BookingInterface` instead of `Booking` => https://github.com/kiwicom/graphql/pull/275#issuecomment-406634734

src/booking/dataloaders/SingleBooking.js
```diff
@@ -24,13 +24,16 @@ export default function createSingleBookingLoader() {
 async function batchLoad(
   bookings: $ReadOnlyArray<Args>,
 ): Promise<Array<Object>> {
-  const promises = bookings.map(({ id, simpleToken }: Args) =>
-    fetchFAQ(id, simpleToken),
+  const promises = bookings.map(({ id, authToken }: Args) =>
+    fetchFAQ(id, authToken),
```
jaroslav-kubicek: `fetchFAQ` wut, is this correct naming? 🤔

jalbertsr: fixed, I see FAQs everywhere 🤦‍♂️ 😅

jalbertsr:
@tbergq @jaroslav-kubicek comments fixed, check again plss

tbergq:
Looks good from my POV, I will wait for @jaroslav-kubicek.

PR #295, opened 2018-07-30
Return authToken from bookingInterface

tbergq:


src/booking/types/outputs/BookingInterface.js
```diff
@@ -279,6 +279,10 @@ export const commonFields = {
       return insurancePrices;
     },
   },
+  authToken: {
+    type: GraphQLString,
+    resolve: ({ authToken }: BookingInterfaceData) => authToken,
```
mrtnzlml: Should be:

```diff
-   resolve: ({ authToken }: BookingInterfaceData) => authToken,
+   resolve: ({ authToken }: BookingInterfaceData): string => authToken,
```

Otherwise, Flow is not able to verify this type correctly. See: https://github.com/kiwicom/graphql#output-types

tbergq:
@mrtnzlml 🚀

PR #296, opened 2018-07-31
Query booking timeline with simple token

tbergq:


src/booking/types/outputs/BookingInterface.js
```diff
@@ -271,11 +271,14 @@ export const commonFields = {
   insurancePrices: {
     type: new GraphQLList(GraphQLInsurancePrice),
     resolve: async (
-      { id }: BookingInterfaceData,
+      { id, authToken }: BookingInterfaceData,
       args: Object,
       { dataLoader }: GraphqlContextType,
     ) => {
-      const { insurancePrices } = await dataLoader.booking.load(id);
+      const { insurancePrices } = await dataLoader.singleBooking.load({
```
RobinCsl: What is the benefit of this approach here?

tbergq: That you can query it while logged in with booking id, IATA code, date and e-mail address 😄

tbergq: see kiwicom/mobile#896

src/booking/queries/__tests__/BookingTimeline.test.js
```diff
@@ -213,4 +213,24 @@ describe('single booking timeline query', () => {
...
+              timestamp
+          }
+        }
+      }
+    `;
+
+    expect(
+      await graphql(query, {
+        id: 'Qm9va2luZzoyNzA3MjI04o',
+        authToken: 'c3f29dd0-18a7-4062-9162-b58f4022fc70',
```
michalsanger: Is this token real?

tbergq: that's a good question, I found it in the same test, @RobinCsl ?

RobinCsl: They are coming from the original datasets files, and they were generated by @mrtnzlml about a year ago. So I assumed they were safe to use?

jalbertsr: yes, it is real, is working for me if I try to build a link to kiwi.com MMB or to API `/users/self/bookings/`, even if it's from a TEST account I think we should use fake token for tests.

tbergq: Probably easier to revoke the token

michalsanger: yeah, revoking is the proper way once it leaked to Github it can be mirrored anywhere. 

btw we have more tokens here, run this:
```bash
grep -rni 'src' -e 'token\"\: \"'
```

michalsanger: According to Booking BE guys, only leaked simple tokens were an issue so they revoked them.

PR #302, opened 2018-08-07
Added regex to check Accept-Language in context

### LINKED ISSUE #300: More robust locale parsing in Context
Inputs like this should work:
```
cs-CZ
cz_CZ
en-US,en;q=0.9,sl;q=0.8
```

dinodsaurus:
closes #300


michalsanger:
If you put the logic into function you can write unit tests and that would be a real pleasure to review :-)

dinodsaurus:
Thanks a lot for the review @michalsanger 🕺

PR #305, opened 2018-08-09
Added support for lovercase countries

dinodsaurus:


src/common/services/context/ParseAcceptLanguage.js
```diff
@@ -2,12 +2,12 @@
 
 export default (acceptLanguage: ?string) => {
   let [language, territory] = ['en', 'US'];
-  // Valid acceptLanguage en_US || en-US || en-US,en;q=0.9,sl;q=0.8
-  const localeRegex = /([a-z][a-z])[-_]([A-Z][A-Z])/;
+  // Valid acceptLanguage en_us || en_US || en-US || en-US,en;q=0.9,sl;q=0.8
+  const localeRegex = /([a-z]{2})[-_]([A-Za-z]{2})/;
```
michalsanger: What about just `de` as an input?

dinodsaurus: Oh, now I see that is also supported by the standard - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language will implement that today 😉

dinodsaurus:
I added `US` as a default country when you don't specify it.
Also `cz+CZ` became valid as it checks first two chars, I would say this is a feature, not a bug 😄 
What do you think @michalsanger

michalsanger:
@dinodsaurus  `cz+CZ` why not, but could you add it into test so it's clear it's intentional?

src/common/services/context/__tests__/ParseAcceptLanguage.test.js
```diff
@@ -9,6 +9,11 @@ describe('Parse Accept-Language', () => {
     expect(con).toBe('US');
   });
   it('should return default values for invalid lang format', () => {
+    const [lang, con] = parseAcceptLanguage('czech');
+    expect(lang).toBe('en');
+    expect(con).toBe('US');
+  });
+  it('should return default values for invalid lang format #2', () => {
     const [lang, con] = parseAcceptLanguage('cz+CZ');
```
michalsanger: > Also cz+CZ became valid as it checks first two chars, I would say this is a feature, not a bug

?

dinodsaurus: yep, I realized when I did this "feature" I broke the code, so I basically removed it 🤥

PR #310, opened 2018-08-14
Adding customer support number query

RobinCsl:
Just one question: should this live under the booking queries to make sure it is protected by token, even though right now endpoint has everything public anyway?

Note as well that the endpoint will be deprecated and replaced at some point.


config/application.js
```diff
@@ -28,6 +28,7 @@ export default {
         queryParameters,
       ),
     rates: 'https://api.skypicker.com/rates',
+    configs: 'https://booking-api.skypicker.com/api/v0.1/configs',
```
mrtnzlml: `v0.1/configs/kiwicom` is the endpoint you should be interested in

mrtnzlml: Also, this file is deprecated. Please do not use it.

src/RootQuery.js
```diff
@@ -28,6 +28,7 @@ import { nodeField } from './node/node';
 import AirHelpPlusOffer from './bookingAncillaries/queries/AirHelpPlusOffer';
 import SingleBooking from './booking/queries/SingleBooking';
 import AllDocuments from './documents/queries/AllDocuments';
+import CustomerSupportNumber from './common/queries/CustomerSupportNumber';
```
mrtnzlml: This query should not be in `common` folder since it's related to customer support - not to everything.

src/common/dataloaders/CustomerSupportNumber.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+
+import DataLoader from 'dataloader';
+import idx from 'idx';
+import { get } from '../../common/services/HttpRequest';
+import Config from '../../../config/application';
+
+let customerSupportNumbersCache = null;
```
mrtnzlml: Do we need this memoization?

src/common/datasets/configs.json
```diff
@@ -0,0 +1,21 @@
+{
+  "kiwicom": {
+    "contacts": {
+      "phones": {
+        "enabled": true,
```
mrtnzlml: Shouldn't this `enabled` configuration be used somewhere?

config/application.js
RobinCsl: Could you tell me which one is meant to be used?

src/common/queries/__tests__/__snapshots__/CustomerSupportNumber.test.js.snap
```diff
@@ -0,0 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`customerSupportNumber query should return the customer support number for the locale in the context 1`] = `
+Object {
+  "data": Object {
+    "customerSupportNumber": null,
+  },
+  "errors": Array [
+    [GraphQLError: Data mock not found for path: https://booking-api.skypicker.com/api/v0.1/configs],
```
mrtnzlml: I think this is a mistake.

src/RootQuery.js
RobinCsl: Sure, I will create a customerSupport folder then.

src/common/dataloaders/CustomerSupportNumber.js
RobinCsl: Probably not.

src/common/queries/__tests__/__snapshots__/CustomerSupportNumber.test.js.snap
RobinCsl: Good catch... I should have been more careful and check the generated snapshot.

config/application.js
mrtnzlml: `v0.1/configs/kiwicom` is subset of `v0.1/configs` (it's shorter without other configs)

src/common/datasets/configs.json
RobinCsl: I suppose I could use it, that would make it better on FE/RN as well to not show it if the phones are unavailable for some reason.

config/application.js
RobinCsl: >  this file is deprecated

I was referring to this. I will use the new endpoint you just mentioned.

src/common/datasets/configs.json
mrtnzlml: Yes, I think they can turn the number in/off on demand. But I don't know how does it work actually.

config/application.js
mrtnzlml: Aha, just add it to your data-loader.

RobinCsl:
@mrtnzlml Took into consideration your comments, it is hopefully better now 😃

src/customerSupport/dataloaders/__tests__/CustomerSupportNumber.test.js
```diff
@@ -0,0 +1,53 @@
...
+      'https://booking-api.skypicker.com/api/v0.1/configs/kiwicom',
+    ).replyWithData(CSNumberEnabledDataset);
+  });
+
+  it('should load customerSupportNumberLoader for valid locale territory', async () => {
+    const customerSupportNumberLoader = createCustomerSupportNumberLoader();
+    const customerSupportNumber = await customerSupportNumberLoader.load(
+      'wonderland',
+    );
+    expect(customerSupportNumber).toEqual({
```
mrtnzlml: You can save some time in the future by using `toMatchInlineSnapshot`. But I will happily accept this version as well. 🙂

src/customerSupport/queries/__tests__/CustomerSupportNumber.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+describe('customerSupportNumber query', () => {
+  it('should return the customer support number for the locale in the context', async () => {
+    const customerSupportNumberQuery = `{
+      customerSupportNumber {
+        localeTerritory
+        number
+      }
+    }`;
+    expect(await graphql(customerSupportNumberQuery)).toMatchSnapshot();
```
mrtnzlml: You could use `toMatchInlineSnapshot` even here. It's very convenient for such short snapshots because you can see it directly next to the tested code.

src/customerSupport/queries/__tests__/CustomerSupportNumber.test.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe('customerSupportNumber query', () => {
+  it('should return the customer support number for the locale in the context', async () => {
+    const customerSupportNumberQuery = `{
+      customerSupportNumber {
+        localeTerritory
+        number
+      }
+    }`;
+    // $FlowExpectedError: Missing types for jest 23.x.x https://github.com/flow-typed/flow-typed/issues/2264
```
mrtnzlml: Please update the `flow-typed` types instead (or add this one function manually to the file - that's also fine).

RobinCsl: 🚀

PR #314, opened 2018-08-15
bugfix: "vehicle" field didn't work for booking collection queries

### LINKED ISSUE #313: "Leg.vehicle" field is broken
~"vehicle" field doesn't work if booking is of `BookingReturn` type~

Affected queries - `customerBookings`, `allBookings`

UPDATE: "vehicle" field didn't work for collection queries which fetched list of bookings from `/api/v0.1/users/self/bookings?token=<token>` instead of `/api/v0.1/users/self/bookings/<bookingId>?simple_token=<token>`

jaroslav-kubicek:
closes #313

### DIFF
--- a/scripts/graphql-schema-snapshot.json
+++ b/scripts/graphql-schema-snapshot.json
@@ -5557,18 +5557,6 @@
             "deprecationReason": null
           },
           {
-            "name": "vehicle",
-            "description": null,
-            "args": [],
-            "type": {
-              "kind": "OBJECT",
-              "name": "Vehicle",
-              "ofType": null
-            },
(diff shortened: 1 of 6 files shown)


michalsanger:
Can there be a test?

jaroslav-kubicek:
I think there can be a test. Fixed :) @michalsanger 

Also, I find our datasets used for tests quite messy, the main issue is that we haven't differentiated between bookings from `/bookings?token=<token>` and `/bookings/<bookingId>?simple_token=<simpleToken>` endpoints, so I opened this issue #315 in which I would like to fix it.

PR #317, opened 2018-08-16
Added supported languages for logitravel

dinodsaurus:


src/dynamicPackage/dataloaders/DynamicPackages.js
```diff
@@ -38,6 +38,15 @@ export default class DynamicPackageDataLoader {
     );
     this.locationDataLoader = location;
     this.supportedMarkets = new Set(['ES', 'GB', 'IT', 'DE', 'FR', 'PT', 'IE']);
+    this.supportedLanguages = new Set([
```
michalsanger: How come the `supportedMarkets` is one line? :-)

dinodsaurus: dunno, prettier did some magic here 😄

michalsanger: Would you accept my naming suggestion so Prettier magic output is one line? It would be `supportedLangs` :-)

PR #320, opened 2018-08-20
Add transportation to whitelabeled services

tbergq:


src/booking/types/outputs/services/TransportationService.js
```diff
@@ -0,0 +1,34 @@
...
+      type: GraphQLList(
+        new GraphQLObjectType({
+          name: 'TransportationServiceRelevantLocations',
+          fields: {
+            location: {
+              type: LocationGraphQLType,
+            },
+            whitelabelURL: {
+              type: GraphQLString,
+              resolve: (): string => 'https://www.kiwi.rideways.com/?',
```
mrtnzlml: Is this URL alright? I would expect some parameters here, no?

tbergq: expect for date, there is not much we can do before user makes his selections

src/booking/types/outputs/services/WhitelabeledServices.js
```diff
@@ -108,5 +114,54 @@ export default new GraphQLObjectType({
...
+        } else if (booking.legs) {
+          const locations = await getLocationsFromLegs(
+            booking.legs,
+            dataLoader,
+          );
+          return [departure, ...locations];
+        }
+        return [];
+      },
+    },
```
mrtnzlml: This is what I call a thin proxy without any logic... 😄

tbergq: Whoever said this does not contain any logic 🙈

tbergq:
@mrtnzlml 🚀

src/booking/resolvers/TransportationService.js
```diff
@@ -0,0 +1,66 @@
...
+
+  return locations.map(location => {
+    const leg = legs.find(
+      leg => leg.arrival.where.code === location.locationId,
+    );
+    return {
+      location,
+      date: idx(leg, _ => _.arrival.when.local),
+    };
+  });
```
mrtnzlml: Yes, this is much much more readable - no more await of the promise during reduce... 😊

PR #321, opened 2018-08-21
WhitelabeledServices: Flatten arrays for multicity and return trip

tbergq:


src/booking/types/outputs/services/__tests__/WhitelabeledServices.transportation.test.js
```diff
@@ -76,22 +76,18 @@ Array [
     "date": 1970-01-01T00:00:00.001Z,
     "location": "OSL",
   },
-  Array [
```
tbergq: snaphots are not so good when we don't read them 🙈 

It was trying to tell us this was wrong

mrtnzlml: No, I usually don't read them. These tests are not good.

tbergq: Should I maybe replace the snapshots here, there is not so huge response anyway

mrtnzlml: Yes, you could do that if you want... :)

PR #323, opened 2018-08-21
Add whoIAm to context + Supply X-WHOIAM to KnowledgeBase

### LINKED ISSUE #324: Set `X-WHOIAM` header for calls to knowledgebase API when posting comments
Knowledgebase API sets a certain rate limit per user. Users are identified by IP + user agent. Therefore each instance of this GraphQL proxy would count as one user, and suffer a debilitatingly small rate limit.

To solve this issue, we have to supply an `X-WHOIAM` header which then will serve as identification for the user. Some documentation about this is available here: https://gitlab.skypicker.com/frontend/wiki/blob/master/general/visitor-id.md

kantord:
fixes #324

### DIFF
--- a/src/FAQ/mutations/__tests__/addFAQArticleComment.test.js
+++ b/src/FAQ/mutations/__tests__/addFAQArticleComment.test.js
@@ -42,4 +42,17 @@ describe('FAQArticleComment mutation', () => {
       }),
     ).toMatchSnapshot();
   });
+  it('should provide X-WHOIAM', async () => {
+    const HttpRequest = require('../../../common/services/HttpRequest');
+    const spy = jest.spyOn(HttpRequest, 'post');
+    RestApiMock.onPost(
+      `https://api.skypicker.com/knowledgebase/api/v1/articles/${goodId}/comment`,
+    ).replyWithData({ message: 'yummy indeed' });
+    await graphql(commentMutation, {
+      articleId: globalId,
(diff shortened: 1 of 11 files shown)


kantord:
@jaroslav-kubicek, @vborszeki  can you please help me out with the failing tests? I don't see why they're failing :-/

mrtnzlml:
Can you please always link an issue or explain in the PR or commit what is it about? It's important for GraphQL maintainers and future visitors. Because last commit message (and therefore PR title) is just "Add whoIAm to context" which explains nothing. Thank you.

mrtnzlml:
@kantord I think description in this PR would be enough so you don't have to create an issue post mortem. It's always better to write it directly to the commit description though (you know - `git log`)... :)

kantord:
Whoops. I already have :-) I'll change the commit message too

src/FAQ/mutations/__tests__/addFAQArticleComment.test.js
```diff
@@ -42,4 +42,17 @@ describe('FAQArticleComment mutation', () => {
       }),
     ).toMatchSnapshot();
   });
+  it('should provide X-WHOIAM', async () => {
+    const HttpRequest = require('../../../common/services/HttpRequest');
```
jaroslav-kubicek: is it possible to import it in regular way?

PR #334, opened 2018-08-27
Add availabilityStatus enum to BoardingPass

Radek-Palisa:


src/booking/types/outputs/BoardingPass.js
```diff
@@ -26,7 +29,12 @@ export default new GraphQLObjectType({
       type: GraphQLDate,
       description:
         'The date when the boarding pass will be available for download',
-      reslove: ({ availableAt }) => availableAt,
+      resolve: ({ availableAt }) => availableAt,
+    },
+    availabilityStatus: {
+      type: BoardingPassStatus,
+      description: 'Boarding pass availability status',
+      resolve: boardingPassStatusResolver(),
```
michalsanger: I see one issue here. If the resolver doesn't get what's expected there is no Flow check to let us know about it. So it's safer to use it like this:
https://github.com/kiwicom/graphql/blob/e1488b2c38cc4d3d26b93336d43027c23310b848/src/hotel/types/outputs/HotelCity.js#L38

src/booking/types/enums/BoardingPassStatus.js
```diff
@@ -0,0 +1,42 @@
...
+
+export const boardingPassStatusResolver = () => ({
+  availableAt,
+  boardingPassUrl,
+}: ResolverType): string => {
+  if (boardingPassUrl) {
+    return AVAILABLE;
+  }
+
+  if (availableAt === null) {
```
michalsanger: What if `availableAt` is `undefined`?

src/booking/types/outputs/__tests__/BoardingPass.test.js
```diff
@@ -57,4 +17,40 @@ Object {
...
   });
+
+  it('should return the corresponding availability status', () => {
+    const offsetDaysFromToday = offsetDays => {
+      return new Date(new Date().setDate(new Date().getDate() + offsetDays))
+        .toISOString()
+        .split('T')[0];
+    };
+
+    expect(
```
michalsanger: It's better for readability to create it's own test for every expect.

src/booking/types/enums/BoardingPassStatus.js
Radek-Palisa: If I understood the specs correctly, `null` will be set deliberately to the field when the boarding pass won't be available online and it can be obtained at the airport only. So `undefined` should only happen if there's a server error I suppose? In that case it will return `OTHER`.

michalsanger: OK if it's intentional. 

This should be covered in the unit tests:
> So undefined should only happen if there's a server error I suppose? In that case it will return OTHER.

src/booking/types/outputs/BoardingPass.js
jaroslav-kubicek: there is probably also no need to define `boardingPassStatusResolver` as function and then calling it :)

jaroslav-kubicek:
pls also rebase, cc @Radek-Palisa

src/booking/types/enums/BoardingPassStatus.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+
+import { GraphQLEnumType } from 'graphql';
+import { DateTime } from 'luxon';
+
+const now = DateTime.utc();
```
tbergq: This means that now will be at the time the file starts to exist (code deployed) and not the time when the function is called.

src/booking/types/outputs/__tests__/BoardingPass.test.js
```diff
@@ -1,52 +1,16 @@
...
 import BoardingPass from '../BoardingPass';
 import { evaluateResolver } from '../../../../common/services/TestingTools';
 
+const now = DateTime.utc();
+
+const availabilityStatus = BoardingPass.getFields().availabilityStatus;
+
 describe('Boarding pass', () => {
   it('should have valid fields', () => {
-    expect(BoardingPass.getFields()).toMatchInlineSnapshot(`
```
tbergq: also why was this changed from inline snapshot, to snapshot?

Radek-Palisa: I did this on the instruction of @jaroslav-kubicek. I assume the reason being that it's better to have the snapshot as a seperate file within the snapshot folder rather having the inlined in the test file.

src/booking/types/enums/BoardingPassStatus.js
Radek-Palisa: thanks, that was silly.

Radek-Palisa:
@michalsanger @jaroslav-kubicek @tbergq can you guys have a look? :)

tbergq:
LGTM, please also rebase master into this branch since `This branch is out-of-date with the base branch`

PR #336, opened 2018-09-04
add "passenger" field to booking contact details

jaroslav-kubicek:
TODO: some tests probably

### DIFF
--- a/scripts/graphql-schema-snapshot.json
+++ b/scripts/graphql-schema-snapshot.json
@@ -7172,6 +7172,18 @@
             },
             "isDeprecated": false,
             "deprecationReason": null
+          },
+          {
+            "name": "passenger",
+            "description": "Documentation and information about the passanger.",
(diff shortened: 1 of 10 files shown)


scripts/graphql-schema-snapshot.json
```diff
@@ -7172,6 +7172,18 @@
             },
             "isDeprecated": false,
             "deprecationReason": null
+          },
+          {
+            "name": "passenger",
+            "description": null,
```
kantord: hmm, don't we need `description`?

jaroslav-kubicek: we should, just ain't no time :D

michalsanger:
@kantord  I'm missing a test where is query with the added property and snapshot where I can see it resolves OK.

jalbertsr:
query test with snapshot added @michalsanger quick check pls

PR #338, opened 2018-09-07
Add leg stopover duration

tbergq:


RobinCsl:
Why do all these snapshots for __generateSomeTimelineEvent__ also include the snapshot from the `BookingTimeline.test.js`: 
```
exports[`resolver matches the snapshot 1`] = `
Array [
  Object {
    "arrival": Object {
      "authToken": "token-lol",
      "bid": 2707251,
      "when": Object {
        "local": 2017-09-09T23:30:00.000Z,
```

michalsanger:
@RobinCsl merge once you get answer plz

RobinCsl:
@michalsanger I am not authorized to merge

RobinCsl:
I don't think it is hugely important, but I don't understand why the toMatchSnapshot for DownloadBoardingPass.test.js would include the snapshot from the `BookingTimeline.test.js`. Is it worth investigating?

tbergq:
I could look into it, but it would be on Monday. Still at the airport. Can’t investigate from here

RobinCsl:
Ok, so let's wait for your investigation on Monday and then we can merge (unless it is very important that we merge it now?)

tbergq:
No, there is no rush

tbergq:
@RobinCsl it seems it happend because of this import `import { leg } from '../BookingTimeline.test';`

I extracted booking and leg from the test into another file. Now the snapshots are gone.

RobinCsl:
@tbergq
Interesting behaviour! Thanks for investigating!
Looks like your branch is out-of-date with `master`

PR #339, opened 2018-09-08
Update "upcomingLeg" field to enable filter out legs by guarantee

jaroslav-kubicek:
We would like to ask for most upcoming leg & take guarantee type into account too.

- without argument - just get an upcoming leg to retrieve easily info about nearest flight
- with argument guarantee=KIWICOM - to get nearest flight covered by our guarantee to know whether we should display additional support


jaroslav-kubicek:
Hola @mrtnzlml & @tbergq could I ask you for CR and approval? I need it to get merged today and everyone but you left me today :)

tbergq:
I will have a look at it when I get to the office, unless @mrtnzlml beats me to it.

src/booking/types/outputs/BookingInterface.js
```diff
@@ -307,8 +308,16 @@ export const commonFields = {
...
-      findUpcomingLeg(legs),
+    args: {
+      guarantee: {
+        type: GraphQLCoveredBy,
+        description: 'Filter out only legs with specified guarantee.',
+      },
+    },
+    resolve: (
+      { legs }: BookingInterfaceData,
+      { guarantee }: Object,
```
tbergq: Could you add a better type here?

jaroslav-kubicek: should be fixed

PR #342, opened 2018-09-10
Allow available hotel price stats to query with currency other than EUR

tbergq:


src/hotel/types/outputs/HotelAvailabilityStats.js
```diff
@@ -26,5 +43,14 @@ export default new GraphQLObjectType({
         });
       },
     },
+    maxPrice: {
+      type: GraphQLFloat,
+      resolve: (ancestor, args: Object, { dataLoader }: GraphqlContextType) => {
+        return dataLoader.hotel.priceStats.load({
+          searchParams: ancestor.searchParams,
```
RobinCsl: Don't you want the same as above?
```
searchParams: {
            ...ancestor.searchParams,
            currency: 'EUR',
          },
```

tbergq: No, this is to avoid BC break. We keep the old fields, so current app can still query with EUR, and we can introduce new field to the app without breaking current version.

PR #346, opened 2018-09-17
Added tests for dynamic packages

dinodsaurus:
Thanks a lot for your help @mrtnzlml regarding the mock stuff 😉


src/dynamicPackage/dataloaders/DynamicPackages.js
```diff
@@ -59,10 +59,12 @@ export default class DynamicPackageDataLoader {
           throw new Error(`No hotels returned from Logitravel`);
         }
         // only show hotels with 3 or more stars
-        // and then shuffle the array
-        return Hotels.filter(h => h.Category >= 3)
-          .sort(() => 0.5 - Math.random())
-          .map(hotel => this.createPackage(Flight, hotel));
+        // and then shuffle the array if running test dont shuffle them
```
mrtnzlml: Hi, I do not think this is OK. You should not change the implementation only to make the tests happy. One way how to deal with this is to mock the random function (there are even libraries for it). Another way is to stop using this snapshot testing (which is my recommendation) and test it differently. For example, you could test here the exception, the returned array length, the category filtering. Another test could take care of the `createPackage` method and so on.

PR #348, opened 2018-09-20
Add X-WHOIAM to header in voteFAQArticle mutation

vborszeki:
Previously mentioned in #324, now we also have to add X-WHOIAM to header for voteFAQArticleMutation.
Closes https://github.com/kiwicom/smart-faq/issues/742


mrtnzlml:
@vborszeki Can you maybe do it for everything and add it to our documentation as a feature? This should be just a thin proxy after all.

PR #352, opened 2018-10-01
Deprecate allAvailableHotels query and replace with more performant version

tbergq:
The reason for this PR is that the existing `allAvailableHotels` query is performing really bad. It takes 8-10 seconds to fetch data about 50 hotels. The reason for this is overfetching when loading dataLoader.hotel.cityById. 

This query will just load what we need to show the search results in the listview, and is much more performant.


tbergq:
> are you going to do cleanup and remove the old query after you change your app

Yes, that was the idea, to cleanup when the native app has added the bundle that will use the new query

PR #356, opened 2018-10-02
singleBooking: extend query to support queries without authToken arg

jaroslav-kubicek:
I would like to do something about our usage of deprecated `booking` query in SmartFAQ and finally stop using it. And I found out that it's necessary to have a query that just accepts `databaseId` as arg. 

Since there is already `singleBooking` query, I suggest adjusting this query (alternatively, there could be yet another query that does the exact same thing, just without `authToken` arg, but I believe it's not necessary.). Opinions welcomed 🙂


src/booking/queries/SingleBooking.js
```diff
@@ -17,7 +17,7 @@ export default {
         'You should use "node" query if you want to use opaque ID.',
     },
     authToken: {
-      type: new GraphQLNonNull(GraphQLString),
+      type: GraphQLString,
```
michalsanger: Why was `GraphQLNonNull` removed?

jaroslav-kubicek: @michalsanger that is what I suggest:

`singleBooking(id, authToken)` - for fetching booking with simple token
`singleBooking(id)` - for fetching booking with access token (sent in header as in other queries)

michalsanger: Oh, I'm blind, my mistake I thought you removed non-null from the ID

